PR #17, opened 2022-07-27
feat:  output tag as placeholder for error message

edmundhung:
## Context

This was a thought I came up with recently. As most of the form will be populating the error message somewhere next to the corresponding field only, while the content of the error doesn't matter. For example:

```tsx
export default function SignupFieldset() {
  const ref = useRef();
  const { email, password, confirm } = useFieldset(ref);

  return (
    <fieldset ref={ref}>
      <label>
        Email:
        <input type="email" name="email" />
        <p>{email.error}</p>
      </label>
      <label>
        Password:
        <input type="password" name="password" />
        <p>{password.error}</p>
      </label>
      <label>
        Confirm Password:
        <input type="password" name="confirm" />
        <p>{confirm.error}</p>
      </label>
      <button type="submit">Sign up</button>
    </fieldset>
  );
}
```

By introducing a placeholder tag, e.g. `<output />` tag, conform can populate the error directly on it without going through react. Imagine rewriting the above example with this:

```tsx
export default function SignupForm() {
  const ref = useRef();

  return (
    <fieldset ref={ref}>
      <label>
        Email:
        <input type="email" name="email" />
        <output name="email" />
      </label>
      <label>
        Password:
        <input type="password" name="password" />
        <output name="password" />
      </label>
      <label>
        Confirm Password:
        <input type="password" name="confirm" />
        <output name="confirm" />
      </label>
      <button type="submit">Sign up</button>
    </fieldset>
  );
}
```

### Notes

- To make this effective, there must be a way for user to opt-in to the existing behaviour which populates the latest error with React when needed. (e.g. by tracking if the users is accessing the error field through a proxy)
- There will be no way to translate the message with the placeholder tag. The use of placeholder tag might only be useful for simple form which performance is not a concern.

### DIFF
--- a/examples/remix/app/helpers.ts
+++ b/examples/remix/app/helpers.ts
@@ -11,14 +11,10 @@ export const styles = {
         peer mt-1 block w-full px-3 py-2 bg-white border
         border-slate-300 rounded-md text-sm shadow-sm 
         focus:outline-none focus:border-indigo-500 focus:ring-1 focus:ring-indigo-500
+        touched:invalid:border-pink-500 touched:invalid:text-pink-600
+        focus:touched:invalid:border-pink-500 focus:touched:invalid:ring-pink-500
     `,
-	invalidInput: `
-        peer mt-1 block w-full px-3 py-2 bg-white border
-        border-slate-300 rounded-md text-sm shadow-sm 
-        focus:outline-none focus:border-indigo-500 focus:ring-1 focus:ring-indigo-500
-        invalid:border-pink-500 invalid:text-pink-600
-        focus:invalid:border-pink-500 focus:invalid:ring-pink-500
-    `,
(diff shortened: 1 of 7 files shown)


danestves:
Any updates on this 👀

edmundhung:
> Any updates on this eyes ?

Not yet. What is your usecase? Now that Conform supports multiple errors, I see more cases that are hard to be supported. I think there is also accessibility concern with output tag  :thinking:

danestves:
> > Any updates on this eyes ?
> 
> Not yet. What is your usecase? Now that Conform supports multiple errors, I see more cases that are hard to be supported. I think there is also accessibility concern with output tag 🤔

Didn't see the new release, that's wonderful! I'm going to play with that to see how to send the array of errors and render it

PR #36, opened 2022-09-21
Add comment about button values

brandonpittman:
I the button value example, it's not clear where the button value is until you look at the Codesandbox, so this commit adds an inline comment in the HTML example.


edmundhung (approved):

edmundhung:
Thanks @brandonpittman. I am actually looking for a better real life examples to showcases this as well. Let me know if you have any ideas.

PR #43, opened 2022-10-22
feat!: simplify validation api

edmundhung:
# Context

This was suppose to be updating docs only. But after trying some ideas, I think this makes the whole validation flow easier to understand. I will create another PR to finalize the documentation.


edmundhung:
Update: I will spend one last day digesting all changes. Tomorrow will be the release day 📦

PR #61, opened 2022-12-20
feat(conform-dom,conform-react): file upload support

edmundhung:
> WIP. This might be split to several PRs later

This PR introduces a new naming convention to denote repeated field names by using the array notation without number, i.e. `[]`. This is particular useful for file input with multiple set to `true`.

This is achieved by an additional mapping logic that look up input by mapping name path on the validation error to the input based on its order on the DOM. e.g. Conform will look up an input with `files[0]` first, and if it is not available, search for the first input with name `file[]`.

More details will be provided later.

### DIFF
--- a/examples/chakra-ui/src/App.tsx
+++ b/examples/chakra-ui/src/App.tsx
@@ -1,9 +1,5 @@
-import {
-	type FieldConfig,
-	useForm,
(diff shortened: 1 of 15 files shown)


edmundhung:
Superseded by #72

Issue #44, opened 2022-10-26
Automatically apply HTML validation attributes based on zod schema


kentcdodds:
Is it possible to have the `conform.input` utility auto-apply HTML validation attributes based on the schema?

kentcdodds:
Along with this, it would be cool if you could give me a `conform.label` so I don't have to manually set the `id` and `htmlFor` to be the same between the label and input.

Something similar to remix-validity-state: https://github.com/brophdawg11/remix-validity-state#usevalidatedinput

Also, note the use of "Prop Getters" over "Prop Collections": https://kentcdodds.com/blog/how-to-give-rendering-control-to-users-with-prop-getters (just a suggestion).

Thanks!

edmundhung:
Hi @kentcdodds, welcome!

> Is it possible to have the `conform.input` utility auto-apply HTML validation attributes based on the schema?

Yes. The `conform.input(config)` helper can already apply the validation attributes if it knows about the constraint. For now, this requires some manual setup: 

```tsx
// This API is only available only on pre-release (v0.4.0) due to some API changes lately.
import { getFieldsetConstraint } from '@conform-to/zod';

function LoginForm() {
  const { email, password } = useFieldset(ref, {
    constraint: getFieldsetConstraint(schema),
  });

  // ...
}
```

This is not documented anywhere because of 2 major issues:
1. Zod schema is almost not analyzable. The helpers mentioned above is just *best effort* as this only works if you have the most basic schema, i.e. `z.object({ email: z.string() })`. But once you try to use some more advance features, e.g. pick / discriminatedUnion, The inner shape definition will be almost impossible to retrieve. The validations is also not directly comparable with the browser one. 
2. Having the validation attributes setup is just the first step. Nest you will wonder how to get the error message from the schema and populate them properly. There are also concerns with [each validation attributes is supported by a subset of the input only](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Constraint_validation#validation-related_attributes). 

> Along with this, it would be cool if you could give me a `conform.label` so I don't have to manually set the `id` and `htmlFor` to be the same between the label and input.

Do you mean an auto-generated `id`? I agree this is handy as I wanted to do the same with the form id too. But I notice `remix-validity-state` is generating it based on `useId()` which requires react 18+.. 🤔 

[...]

PR #77, opened 2023-01-07
feat(conform-react): inferring id for aria-attributes 

edmundhung:
As requested on #44, this enables conform to infer id for label, input and error elements.

As Conform enforce unique name on the dom, it generates:
- Id of the input as `${formId}-${fieldName}`.
- Id of the error as `${formId}-${fieldName}-error`.

This is an opt-in feature. All you need is to provide a form id to the `useForm` hook
```tsx
import {
    useForm,
    useFieldset,
    conform,
} from '@conform-to/react';

export default function LoginForm() {
    const id = useId(); // Only available on React 18, or provide a hardcoded id.
    const form = useForm({ id });
    const { email, password } = useFieldset(form.ref, form.config);

    return (
        <form {...form.props}>
            <div>
                <label htmlFor={email.config.id}>Email</label>
                {/* conform.input() will also return `id`, `aria-invalid` and `aria-describedby` now */}
                <input {...conform.input(email.config, { type: 'email' })} />
                <div id={email.config.errorId}>{email.error}</div>
            </div>
            <div>
                <label htmlFor={password.config.id}>Password</label>
                <input {...conform.input(password.config, { type: 'password' })} />
                <div id={password.config.errorId}>{password.error}</div>
            </div>
            <button type="submit">Login</button>
        </form>
    );
}
```


brandonpittman:
This looks good. Having to pass an if feels a little boilerplatey, but unless you want to require React 18, it’s probably the only good option.

edmundhung:
> This looks good. Having to pass an ID feels a little boilerplatey, but unless you want to require React 18, it’s probably the only good option.

I can also call useId() if [the API is available](https://github.com/reach/reach-ui/blob/dev/packages/auto-id/src/reach-auto-id.ts#L70-L72). But this might become a bit tricky as the behaviour is different based on your react version and it will become a breaking change for react 18 user, if you are always setting a custom id on your form:

```tsx
export default function LoginForm() {
    const form = useForm();

    return (
        <form id="something" {...form.props}>
            {/* something will be overridden by the `id` generated by the generated Id */}
        </form>
    );
}
```

PR #89, opened 2023-01-19
feat(conform-react): validateConstraint helper

edmundhung:
> This feature is designed for client-only app, e.g. react-router

The error message is either the key of the corresponding [ValidityState](https://developer.mozilla.org/en-US/docs/Web/API/ValidityState) or the custom constraint name.

```tsx
import { useForm, validateConstraint } from '@conform-to/react';
import { Form } from 'react-router-dom';

export default function SignupForm() {
    const [form, { email, password, confirmPassword }] = useForm({
        onValidate(context) {
            // This enables validating each field based on the validity state and custom cosntraint if defined 
            return validateConstraint(context, {
                // Define custom constraint 
                match(value, { formData, attributeValue }) {
                    // Check if the value of the field match the value of another field
                    return value === formData.get(attributeValue);
                },
            });
        }
    });

    return (
        <Form method="post" {...form.props}>
            <div>
                <label>Email</label>
                <input
                    name="email"
                    type="email"
                    required
                    pattern="[^@]+@[^@]+\\.[^@]+"
                />
                {email.error === 'valueMissing' ? (
                    <div>Email is required</div>
                ) : email.error === 'typeMismatch' || email.error === 'patternMismatch' ? (
                    <div>Email is invalid</div>
                ) : (
                    <div>{email.error}</div>
                )}
            </div>
            <div>
                <label>Password</label>
                <input
                    name="password"
                    type="password"
                    required
                />
                {password.error === 'valueMissing' ? (
                    <div>Password is required</div>
                ) : (
                    <div>{password.error}</div>
                )}
            </div>
            <div>
                <label>Confirm Password</label>
                <input
                    name="confirmPassword"
                    type="password"
                    required
                    data-constraint-match="password"
                />
                {confirmPassword.error === 'valueMissing' ? (
                    <div>Confirm Password is required</div>
                ) : confirmPassword.error === 'match' ? (
                    <div>Password does not match</div>
                ) : (
                    <div>{confirmPassword.error}</div>
                )}
            </div>
            <button>Signup</button>
        </Form>
    );
}
```


edmundhung:
It might be better to use the attribute name as error message instead of the key on the ValidityState, like this:

```tsx
<div>
    <label>Email</label>
    <input
        name="email"
        type="email"
        required
        pattern="[^@]+@[^@]+\\.[^@]+"
    />
    {email.error === 'required' ? (
        <div>Email is required</div>
    ) : email.error === 'type' || email.error === 'pattern' ? (
        <div>Email is invalid</div>
    ) : (
        <div>{email.error}</div>
    )}
</div>
```

brandonpittman:
Finally made sense of the example. I like the idea—even if I’ll never use it. 😊

edmundhung:
> Finally made sense of the example. I like the idea—even if I’ll never use it. blush

You might use it 😉 I got an idea integrating this with zod, which means you will be able to run zod on server and validate with the native constraint on the client. There's a bit of work to make it happen but if it works out, this will be the main goal for v0.6.

brandonpittman:
> run zod on server and validate with the native constraint on the client

That would be great for a lot of forms.

PR #95, opened 2023-02-09
feat(conform-react)!: simplify validation setup

edmundhung:
> This PR includes breaking changes

This PR remove the default validation logic and replace the `mode` config with a new approach on how Conform decide the validation mode:  

- if `onValidate` is undefined, conform will validate with the server
- If `onValidate` is defined, then it will validate with client validation only
- If async validation is needed, a new `shouldServerValidate` config can be used 

```tsx
export default function Example() {
    const state = useActionData();
    const [form, fieldset] = useForm({
        state,

        // This enable client validation 
        onValidate({ formData }) {
            return parse(formData, { schema });
        },

        // This enable async validation for a specific field
        shouldServerValidate({ submission, defaultShouldValidate }) {
            switch (submission.intent) {
                case 'validate/email':
                    return !hasError(submission.error, 'email');
                default:
                    return defaultShouldValidate;
            }
        },
    });

    // ...
}
```

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -150,7 +150,6 @@ export default function LoginForm() {
   // including the error and config of each field
   const [form, { email, password }] = useForm({
     initialReport: 'onBlur',
-    mode: 'server-validation',
     state: result,
   });
 
--- a/docs/validation.md
+++ b/docs/validation.md
@@ -96,9 +96,6 @@ export default function Signup() {
   // Last submission returned by the server
   const state = useActionData<typeof action>();
   const [form] = useForm<SignupForm>({
-    // Enable server validation mode
-    mode: 'server-validation',
-
     // Begin validating on blur
     initialReport: 'onBlur',
 
@@ -167,16 +164,6 @@ export async function action({ request }: ActionArgs) {
 export default function Signup() {
   const state = useActionData();
   const [form] = useForm({
-    /**
-     * Changing the mode to `client-only` as the client
-     * validation does exactly the same checks as the
-     * server now. There is no need to confirm with the
-     * server again.
-     *
-     * This can be omitted directly as it is the default mode
-     */
-    mode: 'client-only',
-
     state,
 
     // Setup client validation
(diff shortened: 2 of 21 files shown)


brandonpittman:
If any of these cases is `true`, the submission goes to the server and the server runs as normal? The intent that it passes on has no bearing on what the server then does, correct?

edmundhung:
> If any of these cases is `true`, the submission goes to the server and the server runs as normal?

If onValidate is provided without shouldServerValidate, conform will assume client validation would be enough and so it will not send out any submissions with validate intent.

> The intent that it passes on has no bearing on what the server then does, correct?

Yes. This is the same as before. It might look unfamiliar as submission.type is merged with submission.intent.

edmundhung:
The naming is probably confusing with shouldServerValdiate as it is called for any intent (not only validate one)

Conform has a different default for it based on the validation mode.

If onValidate is provided,
shouldServerValidate will be true when intent is not list or validate related

If onValidate is undefined, shouldServerValdiate will be true by default

Any suggestions for a proper name of shouldServerValdiate?🤔

brandonpittman:
shouldPassthrough?

edmundhung:
Renamed to `shouldSubmissionPassthrough`. The name is probably too long. Gonna merge this now and revisit all new names before shipping v0.6. 😅

brandonpittman:
Is `shouldSubmit` appropriate?

edmundhung:
> Is `shouldSubmit` appropriate?

The current `shouldSubmissionPassthrough` is just a simple gate with default value to remove the boilerplate, so calling it as `shouldSubmit` is quite accurate but also becomes less clear on its purpose. People might wonder: When should they use `shouldSubmissionPassthrough` and `onSubmit`?

I actually prefer the original name `shouldServerValidate` quite a lot because of this. Maybe I should adjust ot so that it only get called when the submission intent is `validate` related. For the rest of the case, people should just use `onSubmit` and `event.preventDefault()`.

brandonpittman:
> When should they use shouldSubmissionPassthrough and onSubmit?

It does sound confusing. 

> Maybe I should adjust it so that it only get called when the submission intent is validate related. For the rest of the case, people should just call event.preventDefault() on onSubmit.

That sounds reasonable.

PR #96, opened 2023-02-11
feat!: multiple errors support

edmundhung:
> This PR includes breaking change

I still have a few more changes planned for the `parse` function. This PR introduces a new `resolve` option with a new error structure to support multiple errors:

Here is how you will validate a simple signup form with custom validation:
```tsx
import { parse } from '@conform-to/react';

export async function action({ request }: ActionArgs) {
    const submission = parse(formData, {
        resolve({ email, password, confirmPassword }) {
            // Instead of an array, error is now a simple dictionary
            // The key should be the path while the value would be the corresponding message
            const error: Record<string, string | string[]> = {};

            if (!email) {
                error.email = 'Email is required';
            } else if (!email.includes('@')) {
                error.email = 'Email is invalid';
            }

            if (!password) {
                error.password = 'Password is required';
            }

            if (!confirmPassword) {
                error.confirmPassword = 'Confirm password is required';
            } else if (confirmPassword !== password) {
                error.confirmPassword = 'Password does not match';
            }

            // You can either return an error or a value
            if (error.email || error.password || error.confirmPassword) {
                return { error };
            }

            // You should return the value only if no error 
            return {
                value: {
                    email,
                    password,
                    confirmPassword,
                },
            };
        },
    });

    // You can confirm if there is any error by checking if `submission.value` exists  
    if (!submission.value || submission.intent !== 'submit') {
        return json(submission);
    }

    // ...
}
```

To support multiple error, just assign an array to the error instead of a string:
```tsx
export async function action({ request }: ActionArgs) {
    const submission = parse(formData, {
        resolve({ email, password, confirmPassword }) {
            const error: Record<string, string | string[]> = {};

            // ...

            if (!password) {
                error.password = 'Password is required';
            } else {
                const passwordErrors: string[] = [];
 
                if (password.length < 10) {
                    passwordErrors.push('The password should have minimum 10 characters');
                }

                if (password.toLowerCase() === password) {
                    passwordErrors.push('The password should have at least 1 uppercase character');
                }

                if (password.toUpperCase() === password) {
                    passwordErrors.push('The password should have at least 1 lowercase character');
                }

                // Assign an array of errors instead
                error.password = passwordErrors;
            }

            // ...
        },
    });

    // ...
}
```

You can achieve the same functionality with zod/yup by specifing the `acceptMultipleErrors()` config:
```tsx
import { parse } from '@conform-to/zod';
import { z } from 'zod';

export async function action({ request }: ActionArgs) {
    const submission = parse(formData, {
        schema: z.object({
            // ...
            password: z
                .string()
                .min(10, 'The password should have minimum 10 characters')
                .refine(password => password.toLowerCase() === password, 'The password should have at least 1 uppercase character')
                .refine(password => password.toUpperCase() === password, 'The password should have at least 1 lowercase character')
        }),

        // Default to false if not specified
        acceptMultipleErrors(name) {
            return name === 'password';
        }
    });

    // ...
}
```

You can then access all the errors through `.errors` instead of `.error`. You can always access both `.error` and `.errors` regardless if you have sent multiple errors back. It will just be the first error / an array of 1 item instead: 
```tsx
import { conform, useForm } from '@conform-to/react';

export default function Example() {
    const state = useActionData();
    const [form, { password }] = useForm({
        state,
    });

    return (
        <Form {...form.props}>
            { /* ... */ }
            <div>
                <label>Password</label>
                <input {...conform.input(password.config, { type: 'password' })} />
                <ul>
                    {password.errors?.map((error, i) => (
                        <li key={i}>{error}</li>
                    ))}
                </ul>
            </div>
            { /* ... */ }
        </Form>

    )
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,13 +27,13 @@ function parseLoginForm(formData: FormData) {
   const submission = parse(formData);
 
   if (!submission.payload.email) {
-    submission.error.push(['email', 'Email is required']);
+    submission.error.email = 'Email is required';
   } else if (!email.includes('@')) {
-    submission.error.push(['email', 'Email is invalid']);
+    submission.error.email = 'Email is invalid';
   }
 
   if (!password) {
-    submission.error.push(['password', 'Password is required']);
+    submission.error.password = 'Password is required';
   }
 
   return submission;
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -113,13 +113,13 @@ export async function action({ request }: ActionArgs) {
   // The value will now be available as `submission.payload`
   if (!submission.payload.email) {
     // Define the error as key-value pair instead
-    submission.error.push(['email', 'Email is required']);
+    submission.error.email = 'Email is required';
   } else if (!email.includes('@')) {
-    submission.error.push(['email', 'Email is invalid']);
+    submission.error.email = 'Email is invalid';
   }
 
   if (!password) {
-    submission.error.push(['password', 'Password is required']);
+    submission.error.password = 'Password is required';
   }
 
   // Just check if any error exists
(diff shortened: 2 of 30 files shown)


brandonpittman:
I like this a lot. Would get rid of a lot of my problem where I need to collect errors at the form level. 🚀

PR #98, opened 2023-02-12
feat: shouldValidate config

edmundhung:
This is a hard one. I have been going back and forth on how to handle this properly for a while.

## Background
When you are validating a form, especially with a schema validation library like Zod, you are always validating the whole form at once. However, this get a bit tricky with server validation which includes async validation.

In order to minimize the time taken for validation, conform allows you to optimize it by checking the submission intent with the `shouldValidate` helper to identify if validation for a particular field is actually needed. But this also introduced another problem: Conform cannot reset the error of a particular field if no error is returned unless it **should be validated** (i.e. `shouldValidate` returns true).

The `shouldValidate` helper works with a simple assumption: only the field triggered the validation should be validated. This works fine for most of the form, except when a dependent field is needed in which more than the field triggered needs to be validated as well. 

## Solutions

There are two solutions I thought of:

### 1. Customize it with the `useForm` hook

This ideas requires providing two new custom functions:
- `shouldClientValidate(intent: string, name: string): boolean`
- `shouldServerValdiate(intent: string, name: string): boolean`

These functions works like `shouldValidate` by default and could be customize to handle dependent fields if needed.

Pros: Both custom functions are easy to learn and use.
Cons: You either have to duplicate the `shouldServerValdiate` logic on the server or share it for on-demand validation setup. 

### 2. Includes field names that should be validated as part of the submission

By including the field names, Conform can easily identify which errors could be trusted. This will be set during the `parse` step.

```tsx
export async function action({ request }) {
    const submission = parse(formData, {
        schema: ({ shouldValidate }) => z
            .object({
                // ...
            })
            .refine(value => {
                // You can do on demand validation with
                // the result of `shouldBeValidated`
                if (shouldValidate('password')) {
                    // ...
                }

                // ...
            }),
        shouldBeValidated({ intent, defaultValidated }) {
            if (intent === 'validate/password') {
                // Both fields should be validated together 
                return ['password', 'confirm-password'];
            }
    
            return defaultValidated;
        },
    });

    // This will log ['password', 'confirm-password'] when validating the password field
    console.log(submission.validated);

    // ...
}
```

Pros: By knowing which fields should be validated, we can minimize the data sent back to client by removing error that are not supposed to be validated.
Cons: The custom function is much harder to learn. It is also complicated to find out all the field names especially when dealing with list field, which might need to loop through the parsed result to see how many items in a list. 

## Status

This PR implemented solution 2. But I start wonder if solution 1 is better.

### DIFF
--- a/docs/validation.md
+++ b/docs/validation.md
@@ -224,20 +224,19 @@ export default function Signup() {
 
 ### Validate on-demand
 
-Some validation rules could be expensive especially when they require querying from database or 3rd party services. This can be minimized by checking the submission type and intent, or using the `shouldValidate()` helper.
+Some validation rules could be expensive especially when they require querying from database or 3rd party services. This can be minimized with the `shouldValidate()` helper.
 
 ```tsx
-import { shouldValidate } from '@conform-to/react';
 import { parse } from '@conform-to/zod';
 
 export async function action({ request }: ActionArgs) {
   const formData = await request.formData();
   const submission = parse(formData, {
-    schema: (intent) =>
+    schema: ({ shouldValidate }) =>
       schema.refine(
         async ({ username }) => {
           // Continue checking only if necessary
-          if (!shouldValidate(intent, 'username')) {
+          if (!shouldValidate('username')) {
             return true;
           }
 
--- a/examples/async-validation/app/routes/index.tsx
+++ b/examples/async-validation/app/routes/index.tsx
@@ -1,4 +1,4 @@
-import { conform, useForm, hasError, shouldValidate } from '@conform-to/react';
+import { conform, useForm } from '@conform-to/react';
 import { parse } from '@conform-to/zod';
 import type { ActionArgs } from '@remix-run/node';
 import { json } from '@remix-run/node';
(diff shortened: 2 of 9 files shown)


brandonpittman:
Option 2 seems more flexible to me. I'm sure there would be times where I might want to customize what's going to get validated _at validation time_.

edmundhung:
I am still unsure if solution 2 is a good idea.

In my day job, one of the most complicated form includes a query builder, which has a recursive and nested list that involves both dependent fields and async validation. While it is possible to handle with solution 2, I cant imagine how messy it is gonna be. While solution 1 would be way much simpler as I only need to do some pattern matching based on the name. Having said that, the responsibility does feel more belongs to the parser itself (solution 2). It doesn't feel right to teach the form how to interpret the submission...

I am gonna leave this PR on hold for now as both solution are non breaking changes. I am planning to cut a release before the weekend and I still need to refine some of the previous changes and update all the docs.

brandonpittman:
I don't think either is _bad_. But doing it in the parser will help with conditional logic.

edmundhung:
I found an alternative approach that eliminate the problem mentioned here 😄 I will write up a PR once I finalize the API tomorrow.

PR #146, opened 2023-05-19
Resolve #145: Change initialError, descriptionId, and errorId to all-lowercase

### LINKED ISSUE #145: Conform passes in non-recognized props to <input> fields
When using Conform within [Remix Epic Stack](https://github.com/epicweb-dev/epic-stack/), I receive warnings for three different props that Conform passes in. 

```
Warning: React does not recognize the `initialError` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `initialerror` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
```

This same error is repeated for `descriptionId` and `errorId`.

I believe the solution is to just rename those props to be 100% lowercase, but I'd love to understand if there's a particular reason behind this naming scheme.

lime517:
This resolves [Issue-145](https://github.com/edmundhung/conform/issues/145) by renaming all instances of `initialError`, `descriptionId`, and `errorId` to all lowercase. 

It's also worth mentioning the [HTML Spec](https://html.spec.whatwg.org/multipage/dom.html#embedding-custom-non-visible-data-with-the-data-*-attributes) says that all attributes get lowercased automatically _anyway_, so having attributes as lowercased at the source better matches that spec. 

If readability needed to be improved, the props could be changed to snake-case?

### DIFF
--- a/docs/accessibility.md
+++ b/docs/accessibility.md
@@ -29,7 +29,7 @@ function Example() {
         <form {...form.props}>
             <label htmlFor={message.id}>
             <input {...conform.input(message)}>
-            <div id={message.errorId} role="alert">
+            <div id={message.errorid} role="alert">
                 {message.error}
             </div>
             <button>Send</button>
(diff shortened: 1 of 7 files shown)


edmundhung:
Closing it now as mentioned on #145. Thanks again for the contribution!

PR #172, opened 2023-06-07
fix(react): ensure forms with server-side validations can be submitted

kentcdodds:
Without this change, my form that has server validations will not submit when the client-side is valid but there are some server validations needed.

I'm running on the latest pre-release of `0.7.0-pre-1`.


edmundhung:
This is handled on #168 already and released with `v0.7.0-pre.2`.

PR #173, opened 2023-06-07
feat(conform-react)!: simplify useInputEvent

edmundhung:
```ts
// Before - always returns a tuple with both ref and control object
const [ref, control] = useInputEvent();

// After - You need to provide a ref yourself
const ref = useRef<HTMLInputElement>(null);
const control = useInputEvent({
  ref,
});

// After - You can also provide a function as ref 
const control = useInputEvent({
  ref: () => document.getElementById('do whatever you want'),
});


brandonpittman:
Can't wait for 1.0.

PR #178, opened 2023-06-13
Update validation.md

jessethomson:
I think the schema variable in this example is there on accident. (The `schema` variable gets added in the next code snippet to show how to share the schema between the client and the server)

Feel free to ignore this PR since I never asked if I could make it. Just noticed this as I was reading through the docs.


edmundhung (approved):
Thanks!

PR #194, opened 2023-06-21
refactor: utilize submission payload for form reset instead

### LINKED ISSUE #191: Reset doesn't work after first submission using new method
Hey,

Really enjoying using conform. Just trying out the new reset method in v0.7.0 and can't get it to work, after the second submission it won't reset.

Here is a minimal example: https://codesandbox.io/p/sandbox/jovial-field-h3qckx?embed=1&file=%2Fapp%2Froot.tsx%3A43%2C18

If you submit a value in the text box it will be cleared the first time, afterwards it won't get cleared if you try another input.

Pretty sure it's something to do with passing in the last submission, and probably a useEffect or something under the hood not being updated with new deps.

edmundhung:
Fix #191

On [v0.7.0](https://github.com/edmundhung/conform/releases/tag/v0.7.0), we introduced an approach to reset the form automatically by tracking when the last submission value is cleared.

```tsx
const actionData = useActionData();
  // Pass the submission only if the action was failed
  // Or, skip sending the submission back on success
  lastSubmission: !actionData?.success ? actionData?.submission : null,
});
```

But there is a pitfall here which I overlooked. If the submission has been successful consecutively, the internal `useEffect` call will not be executed as it is still receiving the same `null` value and thus no reset happens.

This PR use a different approach to trigger form reset:
```tsx

export let action = async ({ request }: ActionArgs) => {
    const submission = parse(formData, { schema });

    if (submission.intent !== 'submit' || !submission.value) {
        return json(submission);
    }

    return json({
        ...submission,
        // Notify the client to reset the form using `null`
        payload: null,
    });
};

export default function Component() {
    const lastSubmission = useActionData<typeof action>();
    const [form, { message }] = useForm({
        // The last submission should be updated regardless the submission is successful or not
        // If the submission payload is empty:
        // 1. the form will be reset automatically
        // 2. the default value of the form will also be reset if the document is reloaded (e.g. nojs)
        lastSubmission,
    })

    // ...
}
```

### DIFF
--- a/packages/conform-react/hooks.ts
+++ b/packages/conform-react/hooks.ts
@@ -71,6 +71,12 @@ export type FieldValue<Schema> = Schema extends Primitive
 	? { [Key in KeysOf<Schema>]?: FieldValue<ResolveType<Schema, Key>> }
 	: any;
 
+type SubmissionResult = {
+	intent: Submission['intent'];
+	payload: Submission['payload'] | null;
+	error: Submission['error'];
+};
+
 export interface FormConfig<
 	Output extends Record<string, any>,
 	Input extends Record<string, any> = Output,
@@ -110,7 +116,7 @@ export interface FormConfig<
 	/**
 	 * An object describing the result of the last submission
 	 */
-	lastSubmission?: Submission;
+	lastSubmission?: SubmissionResult;
 
 	/**
 	 * An object describing the constraint of each field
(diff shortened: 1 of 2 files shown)


edmundhung:
Hi @MoSattler. Yes. It was intentional. But now you asked this and I start questioning myself. 

With the original approach being new (which was introduced only last week and so adoption should be low), I think it was acceptable to break it directly, considering the solution was not working as intentional already. I have also updated the release note on v0.7.0 and mention the approach has changed on v0.7.2.

But, maybe it is still a bad idea to break it 😓 I have created #197 which would make it backward compatible, should I merge it and release a patch?

binajmen:
I love it! Very elegant approach :)

PR #195, opened 2023-06-21
feat: improve types

lifeiscontent:
# What I did

- Removed most definitions of `any`
- Fixed some Typo's
- Fixed some value castings that didn't look like they were happening but the types were expecting them (not sure if this was a correct assumption)

### DIFF
--- a/packages/conform-dom/intent.ts
+++ b/packages/conform-dom/intent.ts
@@ -21,12 +21,7 @@ type ExtractListIntentPayload<Operation, Schema = unknown> = Pretty<
 	>
 >;
 
-/**
- * Helpers to configure an intent button for modifying a list
- *
- * @see https://conform.guide/api/react#list
- */
-export const list = new Proxy<{
+type List = {
 	[Operation in ListIntentPayload['operation']]: {} extends ExtractListIntentPayload<Operation>
 		? <Schema>(
 				name: string,
@@ -36,8 +31,15 @@ export const list = new Proxy<{
 				name: string,
 				payload: ExtractListIntentPayload<Operation, Schema>,
 		  ) => IntentButtonProps;
-}>({} as any, {
-	get(_target, operation: any) {
+};
+
+/**
+ * Helpers to configure an intent button for modifying a list
+ *
+ * @see https://conform.guide/api/react#list
+ */
+export const list = new Proxy({} as List, {
+	get(_target, operation) {
 		return (name: string, payload = {}): IntentButtonProps => ({
 			name: INTENT,
 			value: `list/${JSON.stringify({ name, operation, ...payload })}`,
@@ -94,6 +96,7 @@ export function requestIntent(
 	},
 ): void {
 	if (!form) {
+		// eslint-disable-next-line no-console
 		console.warn('No form element is provided');
 		return;
 	}
@@ -142,15 +145,15 @@ export function parseIntent<Schema>(intent: string):
 }
 
 export function updateList<Schema>(
-	list: Array<Schema>,
+	list: Array<Schema | undefined>,
 	payload: ListIntentPayload<Schema>,
-): Array<Schema> {
+): Array<Schema | undefined> {
 	switch (payload.operation) {
 		case 'prepend':
-			list.unshift(payload.defaultValue as any);
+			list.unshift(payload.defaultValue);
 			break;
 		case 'append':
-			list.push(payload.defaultValue as any);
+			list.push(payload.defaultValue);
 			break;
 		case 'replace':
 			list.splice(payload.index, 1, payload.defaultValue);
--- a/packages/conform-validitystate/index.ts
+++ b/packages/conform-validitystate/index.ts
@@ -105,11 +105,12 @@ export type FormConstraints = Record<
  * };
  * ```
  */
-export type FormatErrorArgs<Schema extends FormConstraints = any> = {
-	input: Input;
-	formData: FormData;
-	value: Partial<InferType<Schema>>;
-};
+export type FormatErrorArgs<Schema extends FormConstraints = FormConstraints> =
+	{
+		input: Input;
+		formData: FormData;
+		value: Partial<InferType<Schema>>;
+	};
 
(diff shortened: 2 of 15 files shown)


packages/conform-validitystate/index.ts
```diff
@@ -481,10 +482,10 @@ export function parse<
...
 	const controlMap = new Map<keyof Constraints, Input>();
 	const errorMap = new Map<keyof Constraints, FormError>();
-	const valueMap = new Map<keyof Constraints, any>();
-	// @ts-expect-error FIXME: handle default error type
-	const format: FormatErrorFunction<Constraints, FormError> =
-		schema.formatError ?? defaultFormatError;
+	const valueMap = new Map<keyof Constraints, unknown>();
+	const format =
+		schema.formatError ??
+		(defaultFormatError as FormatErrorFunction<Constraints, FormError>);
```
lifeiscontent, line 488: this casting type is currently as lie, I saw that you had a FIXME here, did you have some other idea for going about fixing this?

packages/conform-validitystate/index.ts
```diff
@@ -697,7 +698,7 @@ export function parse<
 		});
 	}
 
-	const value = Object.fromEntries(valueMap) as any;
+	const value = Object.fromEntries(valueMap) as InferType<Constraints>;
```
lifeiscontent, line 701: this is also a little bit of a lie, I currently just forced it to whatever the used variable is expecting there's also a case where it expects both a Partial<T> and T but didn't make sense of the full context, any suggestions here would be helpful.

packages/conform-yup/index.ts
```diff
@@ -63,15 +63,15 @@ export function getFieldsetConstraint<Source extends yup.AnyObjectSchema>(
 								case 'min':
 									if (
 										!constraint.min ||
-										constraint.min < Number(test.params?.min)
+										Number(constraint.min) < Number(test.params?.min)
```
lifeiscontent, line 66: I casted these, because the original type said they could be a number or string, let me know if you were expecting something else and I'll update accordingly.

packages/conform-yup/index.ts
```diff
@@ -98,7 +98,7 @@ export function parse<Schema extends yup.AnyObjectSchema>(
 		}: {
 			name: string;
 			intent: string;
-			payload: Record<string, any>;
+			payload: Record<string, unknown>;
```
lifeiscontent, line 101: I changed these all to `unknown` so that it was clear we weren't handling the payload directly in library code and if there needed coersion it will warn in the future about potential improper data handling

packages/conform-zod/index.ts
```diff
@@ -96,36 +96,37 @@ export function getFieldsetConstraint<Source extends z.ZodTypeAny>(
 		'pattern',
 	];
 
-	function resolveFieldsetConstarint<T extends Record<string, any>>(
+	function resolveFieldsetConstraint<T extends Record<string, unknown>>(
 		schema: z.ZodType<T>,
 	): FieldsetConstraint<z.input<Source>> {
 		if (schema instanceof z.ZodObject) {
-			const result: FieldsetConstraint<z.input<Source>> = {};
+			const result: Record<string, FieldConstraint<unknown>> = {};
```
lifeiscontent, line 103: the idea here is to cast it to a type that is currently known within this code block

packages/conform-zod/index.ts
```diff
@@ -96,36 +96,37 @@ export function getFieldsetConstraint<Source extends z.ZodTypeAny>(
...
-			for (const [key, def] of Object.entries(schema.shape)) {
-				// @ts-expect-error
+			for (const [key, def] of Object.entries<z.ZodType<unknown>>(
+				schema.shape,
+			)) {
 				result[key] = inferConstraint(def);
 			}
 
-			return result;
+			return result as FieldsetConstraint<z.input<Source>>;
```
lifeiscontent, line 111: and then assume we did our duediligence and hand back the type the caller is expecting.

packages/conform-zod/index.ts
```diff
@@ -96,36 +96,37 @@ export function getFieldsetConstraint<Source extends z.ZodTypeAny>(
...
 		}
 
 		if (schema instanceof z.ZodEffects) {
-			return resolveFieldsetConstarint(schema.innerType());
+			return resolveFieldsetConstraint(schema.innerType());
```
lifeiscontent, line 115: typo

playground/app/components.tsx
```diff
@@ -17,7 +17,9 @@ interface PlaygroundProps<
 	children: ReactNode;
 }
 
-export function Playground({
+export function Playground<
+	TSubmission extends { error: Record<string, string | string[]> | null },
```
lifeiscontent, line 21: I can pull this out if you want, doesn't really matter.

packages/conform-dom/intent.ts
```diff
@@ -21,12 +21,7 @@ type ExtractListIntentPayload<Operation, Schema = unknown> = Pretty<
 	>
 >;
 
-/**
- * Helpers to configure an intent button for modifying a list
- *
- * @see https://conform.guide/api/react#list
- */
-export const list = new Proxy<{
+type List = {
```
lifeiscontent, line 24: the idea here is this is used as our type contract, while the implementation is casted where needed.

packages/conform-dom/intent.ts
```diff
@@ -142,15 +145,15 @@ export function parseIntent<Schema>(intent: string):
 }
 
 export function updateList<Schema>(
-	list: Array<Schema>,
+	list: Array<Schema | undefined>,
 	payload: ListIntentPayload<Schema>,
-): Array<Schema> {
+): Array<Schema | undefined> {
```
lifeiscontent, line 150: this had to change, there's a possiblity that this doesn't make sense, since I believe there's some optionality happening in the ListIntentPayload type, which might indicate that it might be worth being pulled out or rearanged to make this make sense during actual runtime usage.

PR #200, opened 2023-06-29
feat: multiple checkboxes

lifeiscontent:


playground/app/routes/check-boxes.tsx
```diff
@@ -0,0 +1,111 @@
...
+			<Playground title="Attributes" lastSubmission={lastSubmission}>
+				<Field label="Multiple Choice" config={answers}>
+					<label>
+						<input
+							{...conform.input(answers, {
+								type: 'checkbox',
+								value: 'a',
+								id: `${id}-answers-a`,
+							})}
+						/>
```
edmundhung, line 74: Be aware that multiple inputs with the same name is not supported except for radio buttons. It will result in an array in the final payload and the error will be mapped to a path with the corresponding index (e.g. `answers[0]` instead of the original input name `answers`). Radio button is fine as it enforce only 1 value at the end instead.

packages/conform-react/helpers.ts
```diff
@@ -60,24 +70,37 @@ type InputOptions = BaseOptions &
...
 	);
 
-function getFormControlProps(
-	config: FieldConfig<any>,
+function getFormControlProps<
+	Schema,
+	Tag extends 'input' | 'select' | 'textarea',
+>(
+	config: FieldConfig<Schema>,
+	_tag: Tag,
```
edmundhung, line 87: I do not follow why we need to change the result type of this function. It looks strange to pass an argument only as a type hint.

packages/conform-react/helpers.ts
```diff
@@ -4,45 +4,55 @@ import {
...
-	'aria-hidden'?: boolean;
-}
-
-interface InputProps<Schema> extends FormControlProps {
-	type?: HTMLInputTypeAttribute;
-	minLength?: number;
-	maxLength?: number;
-	min?: Schema extends number ? number : string | number;
-	max?: Schema extends number ? number : string | number;
-	step?: Schema extends number ? number : string | number;
```
edmundhung, line 28: All conform helpers are not only designed for native input element but also trying to be compatible with custom input. Although `min` / `max` support both number (e.g. `type=number`) and string (e.g. `type=datetime-local`), I have seen a component library with a number input expecting number only. This was the reason why I added those condition instead of relying on the types of the native input element.

packages/conform-react/helpers.ts
```diff
@@ -60,24 +70,37 @@ type InputOptions = BaseOptions &
 		| {
 				type: 'checkbox' | 'radio';
 				value?: string;
+				id?: string;
 		  }
 		| {
-				type?: Exclude<HTMLInputTypeAttribute, 'button' | 'submit' | 'hidden'>;
+				type?: Exclude<InputProps['type'], 'button' | 'submit' | 'hidden'>;
 				value?: never;
+				id?: never;
```
edmundhung, line 78: If the idea is to provide a way for user to override the default id, I think we can support it with all input type

edmundhung:
Like the idea! Simple enough yet flexible whenever we want to override the default id.

Issue #192, opened 2023-06-21
Doesn't return errors when used with useFetcher


jeyj0:
I just tried using conform in a remix project using `useFetcher` instead of a plain `Form` and it doesn't seem to be working correctly. I have also been unable to find a guide that explains whether some special steps have to be taken for the `useFetcher` case.

When I submit the form using `Form` with wrong input data, I get an error message returned. When I do the same using `fetcher.Form` I don't. I expect the `fetcher.Form` behavior to be the same as the `Form` behavior.

Below is a minimal example using both `Form`, which works as expected, and `useFetcher`, where errors aren't returned on submission and therefore not shown.

<details><summary>The working version using `Form`</summary>
<p>

```tsx
import { useForm } from "@conform-to/react";
import { parse } from "@conform-to/zod";
import { Form, useActionData } from "@remix-run/react";
import type { ActionArgs } from "@remix-run/server-runtime";
import { redirect } from "@remix-run/server-runtime";
import z from "zod";

const schema = z.object({
  name: z.string().nonempty(),
});

export const action = async (args: ActionArgs) => {
  const formData = await args.request.formData();
  const submission = parse(formData, { schema });
  if (!submission.value || submission.intent !== "submit") {
    return submission;
  }
  throw redirect("/");
};

export default function Component() {
  const lastSubmission = useActionData<typeof action>();
  const [form, { name }] = useForm({ lastSubmission });
  return (
    <Form
      method="POST"
      {...form.props}
      className="max-w-screen-sm my-4 mx-auto flex flex-col gap-4"
    >
      <input {...name} className="bg-slate-50 text-slate-950" />
      <span className="text-red-700">{name.error}</span>
      <button className="bg-teal-400 text-teal-950" type="submit">
        Submit
      </button>
    </Form>
  );
}
```

</p>
</details>

<details><summary>The not-working version using `useFetcher`</summary>
<p>

```tsx
import { useForm } from "@conform-to/react";
import { parse } from "@conform-to/zod";
import { useActionData, useFetcher } from "@remix-run/react";
import type { ActionArgs } from "@remix-run/server-runtime";
import { redirect } from "@remix-run/server-runtime";
import z from "zod";

const schema = z.object({
  name: z.string().nonempty(),
});

export const action = async (args: ActionArgs) => {
  const formData = await args.request.formData();
  const submission = parse(formData, { schema });
  if (!submission.value || submission.intent !== "submit") {
    return submission;
  }
  throw redirect("/");
};

[...]

PR #203, opened 2023-07-03
useFetcher examples

jeyj0:
As mentioned in #192, I added a `useFetcher` variant for all `react-router` and `remix` examples.

The remix examples additionally fall back to `useActionData`, which allows the form to still work great with javascript disabled, I found.

The examples are essentially copy-paste from the non-`useFetcher` versions otherwise.

I also added links to the new routes in the root route.


edmundhung (approved):
Thanks for working on it! I have kept only the login form example as the solution is no different.

PR #220, opened 2023-07-15
Rename configuration.md to complex-structures.md

kentcdodds:
I think this page probably shouldn't be at `/configuration`


edmundhung (approved):
Make sense. Thanks!

Issue #175, opened 2023-06-12
How to deal with number inputs?


jansedlon:
Hi again!

I wonder how to deal with number inputs? As per type="number" input spec, it has an available property call `valueAsNumber`. I know that conform does not coerce anything #107 but if it's a native behaviour, why not do it? I saw checkbox and radio specific code in this library so why not number? Is this something I could help you with?

edmundhung:
Hi @jansedlon 

Really appreciate for the offers! The main issue is the same as the checkbox / radio buttons. Since the formData will serialize everything as string except Files. Our server has no idea if a particular entry comes from a specific input type and coerce it on the server side.

I have been exploring a way to reuse the input attributes for validation and coercions based on the input attributes which [I shared on Remix Conf](https://www.youtube.com/watch?v=Y7G6n3kZC9E) last month. This does not work with the Conform solution yet. I am also unsure whether this should be a standalone validation solution or if it worths integrating with Conform tightly  and potentially figure out a way to enhance zod with similar capabilities.

jansedlon:
Hmmm, you're right, probably solution like `zod-form-data` is good enough for now 🤔

edmundhung:
Closing this for now. Feel free to reopen it if you have any new suggestion over this.

Issue #107, opened 2023-03-07
Handle checkboxes with Zod?


kentcdodds:
Does conform do anything to coerce a checkbox to a boolean when using a Zod schema?

brandonpittman:
I don't believe so.

edmundhung:
Conform does not coerce anything as there is no easy way to determine if the form data was set with a checkbox from the server.

You can set it up as `z.string().transform(value => value === 'on')` or `z.preprocess(value => value === 'on', z.boolean())` based on your preference.

edmundhung:
You can now use `z.boolean()` for required checkbox and `z.boolean().optional()` for optional checkbox from [v0.8.0](https://github.com/edmundhung/conform/releases/tag/v0.8.0).

kentcdodds:
Sweet!

luanfonsecap:
How can i validate if at least one checkbox is checked? Any suggestions?

edmundhung:
@luanfonsecap It sounds like you are validating a list of checkboxes. In this case I will define a string array schema with a [nonEmpty](https://zod.dev/?id=nonempty) check.

binajmen:
> @luanfonsecap It sounds like you are validating a list of checkboxes. In this case I will define a string array schema with a [nonEmpty](https://zod.dev/?id=nonempty) check.

But if you only select one checkbox, it becomes a string instead of an array of one value 🤔

edmundhung:
Yep. It will be a string instead of an array of there is only 1 item. But it should work fine because of the automatic type coercion feature with zod which will transform it to an array before validating it.

timzolleis:
Is there an option to use `z.boolean` inside nested objects? Lets say I have:
```ts
const schema = z.object({
  entries: z.array(z.object({
    secret: z.boolean().optional()
  })
})
```
 Which does not transform the checkboxes to booleans (even though they have the correct name) - im using it with "transform" right now. Is this a know bug or by design?

edmundhung:
@timzolleis It might be related to #391 if you have no other text inputs inside the object.

artursopelnik:
I'm using this, like the following:

```typescript
const NON_EMPTY_MESSAGE = "please check this field"
z.boolean(NON_EMPTY_MESSAGE).refine((val) => val === true, {
    message: NON_EMPTY_MESSAGE
})
```

PR #230, opened 2023-07-18
feat(conform-zod): auto type coercion

### LINKED ISSUE #219: Introspect the Zod schema to make array inputs always arrays
There's this caveat in the form guide:

> Conform will transform the value to an array only when there are more than one files selected. To ensure a consistent data structure, you need to preprocess the data as shown in the snippet.

I would like it very much if Conform were to introspect the Zod schema for this. If I expect something to be an array, turn it into one even if only one value was submitted. I did this in my own version of the parser and it was annoying but it worked well: https://github.com/epicweb-dev/rocket-rental/blob/a85622e0539b84f2ff06f4c9aee472e328f9052e/app/utils/forms.tsx#L340

edmundhung:
Fix #219, #107, #175

This enables auto type coercion by recreating the complete schema with an extra `z.preprocess()` setup on each primitive type. It will strip the empty value with automatic type coercion based on the schema type.

If this is merged, the new `stripEmptyValue` option introduced on #224 is no longer necessary.


kentcdodds:
I'd love to bring in @colinhacks and @JacobWeisenburger to ask their opinion on how bad it is to interpolate the zod schema. It would *drastically* improve things for users of Conform (among other libraries I know), but it's a pseudo-private API that I think would be a shame to have broken.

kentcdodds (approved):
I feel pretty good about the implementation and the way it improves the schemas in the tests is only a small taste of how much nicer this will be :)

JacobWeisenburger:
@kentcdodds, what do you mean by "interpolate the zod schema"? What are you interpolating the schema to?

kentcdodds:
Sorry @JacobWeisenburger, I meant "introspect the zod schema" 🤦‍♂️

Here @edmundhung is looking through the zod schema to determine how to transform it to handle `FormData` nicely.

JacobWeisenburger:
@kentcdodds, as far as I know, it's not bad to use the pseudo-private properties as read only. You will likely have problems if you try to write to them. Does that answer your question?

kentcdodds:
Yep, thanks @JacobWeisenburger.

I say let's do it @edmundhung 👍

PR #234, opened 2023-07-19
docs: update comments for (re)validation

muningis:
Original comment suggests that `"onChange"` is allowed, while it's actually `"onInput"`


edmundhung (approved):
Thank you!

Issue #211, opened 2023-07-14
Add a FORM_ERROR export


kentcdodds:
From the docs:

```ts
if (!user) {
    return json({
      ...submission,
      /**
       * By specifying the error path as '' (root), the message will be
       * treated as a form-level error and populated
       * on the client side as `form.error`
       */
      error: {
        '': 'Oops! Something went wrong.',
      },
    });
  }
```

That's kinda weird right? The fact this needs a big comment explaining what's going on is a problem
So what if we do this instead?

```ts
import {FORM_ERROR} from '@conform-to/react' // <-- export this

if (!user) {
    return json({
      ...submission,
      error: {
        [FORM_ERROR]: 'Oops! Something went wrong.',
      },
    });
  }
```

```ts
export const FORM_ERROR = ''
```

It just makes it much more clear and the empty string can remain an implementation detail.

edmundhung:
Thanks for the suggestion, Kent. 

FYI, you don't need to set the form error manually if you do all your checks within the zod schema:

```tsx
const schema = z.object({ ... });
const submission = parse(formData, {
    schema: schema
        .refine(value => {
            // check everything with zod

        }, {
            message: '.....',
            // Note: You don't need to set the path, it's the default on zod.
            path: '',
        })
});
```

But, if you really need to, how about a helper with different options?

```tsx
return JSON({
    submission: report(submission, {
        // This will merged with existing errors on the submission object
        // with the errors provided here at the top (So the error will be populated first)
        formError: 'Oops! Something went wrong.', // or ['error1', 'error2']

        // Another option I have in mind: To reset the form with PE support
        // This is currently achieved by setting submission.payload to `null` manually
        // e.g. { ...submission, payload: null }
        resetForm: true,
    })
})
```

kentcdodds:
I like the `report` helper. I would recommend the documentation always use that utility even if it doesn't do anything without arguments for two reasons:

1. Reduce the "ways to do the same thing" to just 1, even if it's not always necessary
2. Make it possible to easily adapt things in the future if we decide to add defaults or other things to the `report` utility.

edmundhung:
The `report` helper is already implemented and released on `v0.8.0-pre.0`. But I am a bit annoyed with having one more API to import as you always need it. Would it be better to extend the submission object and make the report helper a part of it?

```tsx
return json({
    submission: submission.report({
        /* ... */
    })
})
```

I am also wondering if it should have a different name to avoid confusion, e.g. `result`

```tsx
[...]

PR #246, opened 2023-07-29
feat(conform-react): integrate report helper in submission

edmundhung:
#211


edmundhung:
This will be landed on v1.0 instead.

Issue #263, opened 2023-08-14
Missing params in refine function


jansedlon:
### Describe the bug and the expected behavior

The `refine` function is missing an optional `params` property. This is used for example for translating messages using [zod-i18n](https://github.com/aiji42/zod-i18n#custom-errors).

### Conform version

v0.8.0

### Steps to Reproduce the Bug or Issue

Not possible to use 
```ts
refine(ctx, {
  when: intent === 'submit' || intent === 'validate/field',
  validate: () => validator(),
  params: {
    i18n: "key"
  }
})
```

jansedlon:
I also noticed that the property `message` is required. In this case it shouldn't be required

edmundhung:
Fixed on [v0.8.1](https://github.com/edmundhung/conform/releases/tag/v0.8.1) thanks to the contribution from @jansedlon!

PR #264, opened 2023-08-14
fix: add params to refine

jansedlon:
#263

### DIFF
--- a/packages/conform-zod/parse.ts
+++ b/packages/conform-zod/parse.ts
@@ -14,6 +14,7 @@ import {
 	type ZodErrorMap,
 	type IssueData,
 	ZodIssueCode,
+	ZodCustomIssue,
 } from 'zod';
 import { enableTypeCoercion } from './coercion';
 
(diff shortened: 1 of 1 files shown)


packages/conform-zod/parse.ts
```diff
@@ -105,18 +106,23 @@ export function refine(
...
+		 */
+		params?: ZodCustomIssue['params'];
 	},
 ): void | Promise<void> {
 	if (typeof options.when !== 'undefined' && !options.when) {
 		ctx.addIssue({
 			code: ZodIssueCode.custom,
 			message: VALIDATION_SKIPPED,
 			path: options.path,
+			params: options.params,
```
edmundhung, line 125: Both `VALIDATION_SKIPPED` and `VALIDATION_UNDEFINED` are custom messages that Conform used internally and they should never be translated. Can we skip the `params` for these two issues? I assume zod-i18n will just ignore them and leave them as is if there is no `params` set?

edmundhung (commented):
Thanks for the PR, @jansedlon! I would love to support zod-i18n but I have a minor question.

packages/conform-zod/parse.ts
jansedlon: The zod-i18n translates it only if you pass it under 18n property, like 
```ts
params: {
  i18n: "translationKey"
}
```

Do you think it's still gonna be an issue?

jansedlon: @edmundhung A week gone by and it finally clicked what you meant 🥲 Yeah, i removed the params for the invalid cases, now it's passed only at the end

edmundhung (approved):
Thank you!

PR #270, opened 2023-08-21
List insert

albert-schilling:
This PR extends the `list` helper with an `insert` (in between) method.

```
{/* To insert a new row with optional defaultValue at a given index */}
<button {...list.insert('name', { defaultValue, index })}>Insert</button>
```


edmundhung:
Thanks @albert-schilling! I am happy with the new changes. But Can you undo the version bump? I will prefer taking care of it separately. 

Now that we have `insert`, I might want to deprecate `append` / `prepend` so there won't be several ways to do the same thing. Perhaps making index optional and default to append. Wdyt?

albert-schilling:
> Thanks @albert-schilling! I am happy with the new changes. But Can you undo the version bump? I will prefer taking care of it separately.
> 
> Now that we have `insert`, I might want to deprecate `append` / `prepend` so there won't be several ways to do the same thing. Perhaps making index optional and default to append. Wdyt?

Thanks for the quick review.
I agree. Keeping the interface as small as sensible is always good. 

I updated the guide and added the deprecation hint, but I did not find a way to add the deprecation warning in the type annotation of the `append` and `prepend` methods. Do you know how to do that?

I also changed all tests to test insert instead of prepend and append. Is that okay with you?

edmundhung:
Thank you! @albert-schilling

PR #272, opened 2023-08-24
fix(ts): set payload to unknown value type

kentcdodds:
The Jsonify utility used by Remix will make a `Record<string, any>` disappear in the type: https://github.com/remix-run/remix/issues/7246

But considering this further, "unknown" is a safer option for this type anyway because it avoids issues with accidentally accessing the values off the submission payload without first checking them. So I do think this change should be made.

### DIFF
--- a/packages/conform-dom/parse.ts
+++ b/packages/conform-dom/parse.ts
@@ -3,7 +3,7 @@ import { INTENT, getIntent, parseIntent, updateList } from './intent.js';
 
 export type Submission<Schema = any> = {
 	intent: string;
-	payload: Record<string, any>;
+	payload: Record<string, unknown>;
 	error: Record<string, string[]>;
 	value?: Schema | null;
 };


MichaelDeBoey:
The TS also suggests to use `unknown` instead of `any`: https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#any

MichaelDeBoey (approved):

danestves:
Can confirm that this fix works with the new `v2` of Remix

| Before | After |
|--------|--------|
|  |  |

edmundhung (approved):
Thank you!

danestves:
Any progress on this? 😬

edmundhung:
I will update the remix fixture to v2 (pre) this weekend to verify the change and make a release. 

I was betting the issue to be resolved upstream soon so I don't need to bump the minor version now (This is a breaking change unfortunately). But it looks like `type-fest` is still waiting for a proper fix and the issue on remix has been de-prioritized as well. 😅

PR #275, opened 2023-08-25
fix: typos in checkbox docs

justintoman:
I noticed some typos when I was copy/pasting this into my app to test it out.


edmundhung (approved):
Thank you!

PR #286, opened 2023-09-06
fix: add missing case in useFieldList

albert-schilling:
This fixes a bug occurring during client validation, that originates from a missing case for the new `list.insert` method in the `useFieldList` hook.


edmundhung (approved):
Thank you! I will add a test later today and then release a patch.

PR #292, opened 2023-09-23
Update hooks.ts lastsubmission to accept null value

zifahm:
In Remix, there may be times when you need to return json(null), which is often what Remix expects. If you return json(undefined), the loaders are not refetched. To resolve this issue, accepting null for "lastsubmission" keeps TypeScript happy and ensures the loader is refetched.

### DIFF
--- a/packages/conform-react/hooks.ts
+++ b/packages/conform-react/hooks.ts
@@ -116,7 +116,7 @@ export interface FormConfig<
 	/**
 	 * An object describing the result of the last submission
 	 */
-	lastSubmission?: SubmissionResult;
+	lastSubmission?: SubmissionResult | null;
 
 	/**
 	 * An object describing the constraint of each field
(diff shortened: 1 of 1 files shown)


edmundhung:
Hi @zifahm, I would like to learn more about your usecase first. Can you tell me more what you are trying to do? That will help me understand your problem and identify if there is a better solution :)

zifahm:
I don't always use "conform" for all of my forms. I might also return "json(null)" frequently when I simply need to execute an action. However, TypeScript's type checking becomes problematic when using "json(null)" instead of "json(undefined)," and strangely, using "json(undefined)" doesn't trigger the loader to refresh for some reason. This is the scenario I'm dealing with.

As for my current workaround, I use the following code snippet:

```typescript
lastSubmission: lastSubmission === null ? undefined : lastSubmission
```

I really don't want to have to do this repeatedly. Therefore, allowing "conform" to accept "null" values would reduce the amount of code I need to write.

edmundhung (approved):
I am fine with the changes. I will do a quick check over the weekend and release it as `0.9.1` if I don't see other issue. Thanks! 👍🏼

PR #294, opened 2023-09-27
docs(complex-structures): improve nested array example

fiws:
* add missing import
* define missing variables


edmundhung (approved):
Thank you!

PR #296, opened 2023-10-01
docs: Add closing </div> tags to HTML snippet

mattmazzola:
I was reading through the docs and based on the indentation, it seems the elements that were supposed to contain the error messages from `confirm` were not closed.


edmundhung (approved):
Thank you!

PR #320, opened 2023-11-01
docs: Update README code example

Littletonconnor:
According to the remix docs, `useActionData` comes from `@remix-run/react` https://remix.run/docs/en/main/hooks/use-action-data


edmundhung (approved):
Thank you!

PR #326, opened 2023-11-10
Pass Output type to Submission in onSubmit

rphlmr:
Hello,

I apply this patch on my projects and i think it can be useful.

Make `context.submission` aware of the `Output` type.

Now we should be able to have intellisense with `onSubmit` `context` param.

Ex:
```ts
onSubmit(event, context) {
   // it has type
    if (context.submission.value?.step === "send-otp") {
        // do something
    }
},
```

### DIFF
--- a/packages/conform-react/hooks.ts
+++ b/packages/conform-react/hooks.ts
@@ -156,7 +156,7 @@ export interface FormConfig<
 		event: FormEvent<HTMLFormElement>,
 		context: {
 			formData: FormData;
-			submission: Submission;
+			submission: Submission<Output>;
 			action: string;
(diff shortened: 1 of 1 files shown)


edmundhung:
Hi @rphlmr, thanks for the PR. I would love to improve the types here but your change is not accurate because there would be default submission object returned even when you didn't provide an `onValidate` handler. 

To fix it, we might need to introduce an additional generic to capture the submission returned by the `onValidate` handler or some other adjustment on this behaviour.

edmundhung:
This should be fixed on v1 as it will not returns a default submission anymore. Thanks!

PR #328, opened 2023-11-11
docs: fix async validation example imports

afmelsaidy:
The async validation example imports `refine` from `@conform-to/react` instead of `@conform-to/zod`


edmundhung (approved):
Thank you!

PR #334, opened 2023-11-17
fix(conform-zod): do not coerce whitespace to 0

### LINKED ISSUE #333: Whitespace chars in number fields are incorrectly coerced to 0 with zod

When using z.number() with @conform-to/zod, whitespace characters, e.g. `" "` are incorrectly coerced to 0 by the enableTypeCoercion fn. The problem is that the JS `Number` function handles whitespace characters as 0. A fix would be to use `parseFloat` instead of `Number` on [L124](https://github.com/edmundhung/conform/blob/main/packages/conform-zod/coercion.ts#L124).

### Conform version

v0.9.0

1. Create a number field which is validated with z.number()
2. Add whitespace characters, e.g. `"   "`
3. Validation passes and the input value is coerced to `0`. Validation should fail instead.

_No response_

### Screenshots or Videos

_No response_

### Additional context

_No response_

marcomuser:
Fixes #333

### DIFF
--- a/docs/validation.md
+++ b/docs/validation.md
@@ -99,7 +99,7 @@ export default function Signup() {
 Here is an example how you can do async validation with zod:
 
 ```tsx
-import { refine } from '@conform-to/react';
+import { refine } from '@conform-to/zod';
 
 // Instead of reusing a schema, let's prepare a schema creator
 function createSchema(constraint?: {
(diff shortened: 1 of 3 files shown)


marcomuser:
While this would fix #333, seems like there are more considerations to handle according to the tests... 🤔

marcomuser:
Updated the PR to keep the Number fn to coerce the input but added an additional check to handle white space characters only user input. I also added a simple test to check for this case. What do you think? :-)

edmundhung (approved):
Thanks you! I think it makes sense to add an extra check for whitespace. This should be good enough for simple usage. If anyone want a more custom parsing logic, they should just use `z.preprocess` or `z.transform()`.

edmundhung:
Sorry for holding up this PR for so long. The change will be available on v1. Thanks again and happy new year!

PR #340, opened 2023-11-20
fix(conform-zod): auto type coercion with discriminated union

### LINKED ISSUE #339: conform's automatic type coercion doesn't work with `z.discriminatedUnion`

`z.discriminatedUnion` should be doing automatic type coercion like a normal schema that just uses a `z.object` but it doesn't seem to do that. Everything works fine if i add `.coerce` to the failing fields in my schema.

### Conform version

0.9.0

```typescript
const schema = z.discriminatedUnion("type", [
  z.object({
    type: z.literal("copy"),
    id: z.number(),
  }),
]);
```

Fails validation with the following error:

```
{
  intent: 'submit',
  payload: { type: 'copy', id: '43' },
  error: { id: [ 'Expected number, received string' ] }
}
```

Form:

```jsx
<Form {...formProps} method="POST">
  <input
    type="hidden"
    name="type"
    value="copy"
  />
  <input
    type="hidden"
    name="id"
    value={id}
  />
  <Button type="submit">Copy</Button>
</Form>
```

_No response_

### Screenshots or Videos

_No response_

### Additional context

_No response_

edmundhung:
Fix #339

### DIFF
--- a/docs/validation.md
+++ b/docs/validation.md
@@ -99,7 +99,7 @@ export default function Signup() {
 Here is an example how you can do async validation with zod:
 
 ```tsx
-import { refine } from '@conform-to/react';
+import { refine } from '@conform-to/zod';
 
 // Instead of reusing a schema, let's prepare a schema creator
 function createSchema(constraint?: {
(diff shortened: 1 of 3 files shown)


edmundhung:
This change is applied in https://github.com/edmundhung/conform/commit/d666c33b8af86aa033aa235682505c760efa353a and should be released with v1.

PR #368, opened 2023-12-29
feat: typebox

lifeiscontent:
This PR introduces a new provider, TypeBox which is more than 4x smaller than Zod and fairly common in Node.js Projects for defining schemas

Not only is Typebox smaller and faster than Zod. It's also more popular.

https://npmtrends.com/@sinclair/typebox-vs-yup-vs-zod


edmundhung:
Thanks for working on this! Although I think yup and zod are still more popular as [jest](https://www.npmjs.com/package/@jest/schemas) depends on typebox internally and contributed most of the download counts for typebox, I am not against adding support to different solutions. But there are a few things I wanna learn more first.

1. How do we add custom validation to typebox? For example, it's common to ask for the password twice in a signup form and validated whether both passwords match. How can we implement this using typebox?
2. Does typebox support async validation? If yes, can we also support it in the integration?
3. How does the coercion work and how does it handle empty string?

Really appreciate all the effort and I wish you a happy new year :)

edmundhung:
Sorry for the wait! I will give it a try this week and see if I have any other concern before getting this merged. 🙌🏼

aleshchynskyi:
👋  Just wanna bring a little attention to this PR.
Was thinking of adding `conform` to my project, but having two type validators doesn't sound attractive. Also I really like TypeBox, so many might benefit from this

patlux:
Any updates? :)

stephen776:
+1 tools like Elysia in the bun world are using Typebox in the same manner that tRPC uses zod. 

Would be nice to opt into typebox in cases where we are already using it elsewhere in the stack.

Issue #371, opened 2024-01-07
docs: missing favicon.com


mattmazzola:
### Describe the bug and the expected behavior

It seems the docs are built using Remix and Markdown docs here: https://github.com/edmundhung/conform/tree/main/guide
Although there isn't an explicit meta entry for the icon, it defaults to /favicon.ico but this file does not exist in the public root

![image](https://github.com/edmundhung/conform/assets/2856501/cf902c77-5dcc-4523-a9a4-c24e9b3d4b1b)

### Conform version

N/A

### Steps to Reproduce the Bug or Issue

View docs

### What browsers are you seeing the problem on?

Microsoft Edge

### Screenshots or Videos

![image](https://github.com/edmundhung/conform/assets/2856501/304bf51c-b6e0-4e59-9cc6-4810847974a2)

edmundhung:
Conform doesn't have an official logo yet unfortunately so we don't have any favicon to show. Do you know if there is a way to stop the browser from requesting it? 😅

PR #379, opened 2024-01-14
Add /favicon.ico and links

mattmazzola:
Create `favicon.ico` using https://favicon.io/favicon-generator

**Settings:**
Font: Ubuntu
Style: Bold
Text Color: D9D9D9
Background Color: 18181B

image

It's not fancy, but it should at least prevent this blank file icon which makes the site look broken
image

Fixes: #371


mattmazzola:
I think someone would need to verify that it actually works by generating or using the guide/docs locally. I only added the files but didn't teset

PR #380, opened 2024-01-14
fix spelling

dmatis:


dmatis:
Replaced by: https://github.com/edmundhung/conform/pull/381

PR #394, opened 2024-01-26
Type fixes

aaronadamsCA:
- Remove `image` as an accepted value for input `type`.
- Add `undefined` as an accepted value for `lastResult`, for improved compatibility with TypeScript's `exactOptionalPropertyTypes`.


edmundhung (approved):
Thanks!

PR #398, opened 2024-01-30
Add support for z.catch coercion

### LINKED ISSUE #397: Compatibility issue with `ZodCatch`

If I add a `.catch()` to a `z.number()` schema, `parseWithZod()` always returns the catch value, even with a valid input that matches the Zod schema.

### Conform version

v1.0.0-rc.1

Minimal reproduction: https://stackblitz.com/edit/remix-run-remix-qxwt5m?file=app%2Froutes%2F_index.tsx

Expected behaviour:

```ts
const schema = z.object({
  page: z.number().int().min(1).safe().catch(1),
});

expect(
  parseWithZod(new URLSearchParams('page=2'), {
    schema,
  })
).toEqual({
  status: 'success',
  payload: {
    page: '2',
  },
  value: {
    page: 2,
  },
  reply: expect.any(Function),
});
```

Actual behaviour: `value` is always equal to `{ page: 1 }`.

_No response_

### Screenshots or Videos

_No response_

### Additional context

The use case here is validating `searchParams`, where unlike `formData` you probably just want to fall back to valid values and ignore invalid ones.

If you remove the `.catch()` from the schema, it works as expected—but of course then you can't fall back to valid values.

aaronadamsCA:
Fixes #397

### DIFF
--- a/packages/conform-zod/coercion.ts
+++ b/packages/conform-zod/coercion.ts
@@ -221,14 +221,10 @@ export function enableTypeCoercion<Schema extends ZodTypeAny>(
 				}),
 			);
 	} else if (type instanceof ZodCatch) {
-		schema = any()
-			.transform((value) => coerceFile(coerceString(value)))
-			.pipe(
-				new ZodCatch({
-					...type._def,
-					innerType: enableTypeCoercion(type.removeCatch(), cache),
-				}),
-			);
+		schema = new ZodCatch({
+			...type._def,
+			innerType: enableTypeCoercion(type.removeCatch(), cache),
+		});
 	} else if (type instanceof ZodIntersection) {
 		schema = new ZodIntersection({
 			...type._def,
--- a/packages/conform-zod/coercion.ts
+++ b/packages/conform-zod/coercion.ts
@@ -25,6 +25,7 @@ import {
 	ZodDefault,
 	lazy,
 	any,
+	ZodCatch,
 } from 'zod';
 
 /**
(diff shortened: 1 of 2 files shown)


tests/conform-zod.spec.ts
```diff
@@ -869,6 +869,53 @@ describe('conform-zod', () => {
...
+					createFormData([
+						['a', ''],
+						['b', ''],
+						['c', ''],
+						['d', ''],
+						['e', emptyFile],
+						['f', ''],
+					]),
+					{ schema },
+				),
```
edmundhung, line 896: Can you add one more assertion just to verify what happens if the formdata is valid? e.g. `a` is `helloeworld` and `b` is `456`.

aaronadamsCA: Done.

packages/conform-zod/coercion.ts
```diff
@@ -219,6 +220,15 @@ export function enableTypeCoercion<Schema extends ZodTypeAny>(
...
 			);
+	} else if (type instanceof ZodCatch) {
+		schema = any()
+			.transform((value) => coerceFile(coerceString(value)))
+			.pipe(
+				new ZodCatch({
+					...type._def,
+					innerType: enableTypeCoercion(type.removeCatch(), cache),
+				}),
+			);
```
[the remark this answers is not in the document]
aaronadamsCA: It does! Thanks for that, as it was I just copied the default block to start with.

edmundhung (approved):
Thanks for working on this! I will probably release it with v1 directly.

PR #402, opened 2024-02-01
docs: fix error property in 'UI library' documentation

marilari88:
With the new version the fieldName.error was renamed to fieldName.errors


edmundhung (approved):
Thanks!

PR #409, opened 2024-02-01
feat: accept generic type in `useInputControl`

bryantbrock:
## Problem
`useInputControl` only passes around a `string` for the type. However, the type could possibly be any type for the given field.

## Solution
Give consumers a way to set the type when using `useInputControl`. Now `control.value` and `control.change` will pass around the generic type instead of simply `string`.

Example:
```ts
const control = useInputControl<boolean>(fields.isActive)
```

### DIFF
--- a/packages/conform-react/integrations.ts
+++ b/packages/conform-react/integrations.ts
@@ -1,9 +1,9 @@
 import { type FieldElement, isFieldElement } from '@conform-to/dom';
 import { type Key, useRef, useState, useMemo, useEffect } from 'react';
 
-export type InputControl = {
-	value: string | undefined;
-	change: (value: string) => void;
+export type InputControl<T> = {
+	value: T | undefined;
+	change: (value: T) => void;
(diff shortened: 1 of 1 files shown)


edmundhung:
Hi @bryantbrock, you should be able to pass a boolean field to `useInputControl` already. Changing the type won't affect the actual type of the `control.value` as they are always based on the serialized form value from the `FormData`. This is definitely a mental shift from the traditional form library but is one of the limitation working with the FormData API.

edmundhung:
Thanks for contributing! If you have any questions, feel free to ask here.

bryantbrock:
Thanks for the reply @edmundhung. This was a bit of a workaround for sure, and I now understand better. I was attempting to build my own version of `useFieldArray` with this, but it didn't work for me. I did come up with a solution using native forms (e.g. multiple inputs with the same name), I'll post a discussion about it later on, it may prove valuable to others.

Issue #401, opened 2024-01-31
Conform v1 useInputControl doesn't populate defaultValues correctly


willhoney7:
### Describe the bug and the expected behavior

If you set a `defaultValue` in the `useForm` hook (like for existing server data) and the field is handled by a `useInputControl` hook instead of a normal input (like say if you have a custom select you want to use), the default value for the field is not sent up in the payload, but it is still returned in the `useInputControl` hook. Check out the codesandbox to best see the issue.

Thanks!

### Conform version

v1.0.0

### Steps to Reproduce the Bug or Issue

1. Go to codesandbox: https://codesandbox.io/p/devbox/loving-elgamal-qt7whx?workspaceId=619a0863-8060-480c-8a4c-13e3599255da
2. Go to todo list example
3. Notice that `current value` of the task is `c` 
4. Click save, notice that the submission errors out
5. Looking at console logs, you can see that the `val` is not included in the payload.

### What browsers are you seeing the problem on?

Chrome

bkspace:
I assume the `useInputControl` exists just as a helper to handle controlled form inputs, but in the example there are no inputs. I believe you just need a hidden form input here for each task, such that when conform reads the formData, the value(s) exists in the payload.

edmundhung:
It's a miss on my side. Right now it's only creating a hidden input for you on the first time `control.change(value)` is called. So you won't have defaultValue set on the formData if you never touch the input. I will push up a fix over the weekend. 

For now, you can render an hidden input manually and `useInputControl` will pick it up instead of inserting another one.

edmundhung:
This should be resolved in [v1.0.1](https://github.com/edmundhung/conform/releases/tag/v1.0.1).

anton-g:
I have an issue that might be related. I use Remix with the EpicStack which comes with checkbox component utilizing `useInputControl` (https://github.com/epicweb-dev/epic-stack/blob/90f4208395ab77605082a578e0e7e2cb0c2b66bc/app/components/forms.tsx#L108).

I use `<CheckboxField/>` together with `getCollectionProps` to render a list of checkboxes where you can select one or many.

If I set the default value in `useForm` so that one of the checkboxes is default checked and:

a) publish the form immediately -> the form is submitted with the correct default value
b) check another checkbox and publish the form -> the form is submitted **without** the default value and only the manually selected value.

I'll try to recreate it in a sandbox as soon as I get some time! Let me know if you want it in a separate issue.

anton-g:
[...]

PR #419, opened 2024-02-04
fix(conform-react): useInputControl should populate initial value and support mult select

edmundhung:
This fixes #389 and #401.

But there is one minor issue with multi select. As the `initialValue` could be `string` instead of `string[]` in one edge case in which user select only one item AND submit the form AND a full document reload happened. This is because conform have no idea this field is an array and only find one entry from the FormData. 😅 

There are a few options:
1. Let user deal with it like how I did on the `headless-ui` example:
```tsx
const control = useInputControl(meta);
// To make it an array...
const value = typeof control.value === 'undefined'
  ? []
  : Array.isArray(control.value)
  ? control.value
  : [control.value];
```
2. Add a new properties to the `control` object which wraps the logic above, e.g. `control.valueAsArray`
3. Help user lie 😅 This is an edge case, especially in combination with `useInputControl` which will requires JS anyway and it's weird to expect full document reload to work well..?

### DIFF
--- a/examples/headless-ui/src/App.tsx
+++ b/examples/headless-ui/src/App.tsx
@@ -152,18 +152,26 @@ function classNames(...classes: Array<string | boolean>): string {
 	return classes.filter(Boolean).join(' ');
 }
 
-function ExampleListBox(props: { name: FieldName<string> }) {
+function ExampleListBox(props: { name: FieldName<string[]> }) {
 	const [meta] = useField(props.name);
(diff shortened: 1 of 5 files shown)


EmilEinarsen:
Thanks for this!
From my perspective, option 1 feels consistent with how the formData would need to be handled. 
Just like how when parsing there's a destinction between `payload` and `value`.

Issue #443, opened 2024-02-10
Form not found errors in v1.0.1


adaboese:
### Describe the bug and the expected behavior

Started to get errors that just say "Form not found"

No indication of what's broken. Refreshing the page fixes the issue most of the time.

### Conform version

v1.0.1

### Steps to Reproduce the Bug or Issue

N/A

### What browsers are you seeing the problem on?

Chrome

edmundhung:
You might have a hydration mismatch that cause the `id` to be different between client and server. The error your referred comes from `useInputControl` which tries to look up the form element on a `useEffect`. But I can see that you are setting the form id with `getFormProps()` correctly. So it's very likely a broader issue.

edmundhung:
I am not sure if this is the issue here. As I assume the components that calls `useInputControl` is also rendered inside the ClientOnly tree. I don't think it run before your form element is rendered.

edmundhung:
If you expect the `referenceArticleUrlsInputControl.change()` to update the textarea, then you should register it with `referenceArticleUrlsInputControl.register()` when using `useControl()`:

```tsx
<Textarea
  disabled={submitting}
  onChange={trackInputValue}
  placeholder="https://..."
  resize="none"
  rows={6}
  ref={referenceArticleUrlsInputControl.register}
  {...getInputProps(fields.referenceArticleUrls, {
    type: 'text',
  })}
/>
```

This basically emulate a change happened on textarea and dispatch the corresponding input / change event just so Conform can revalidate / update the internal state.

Issue #306, opened 2023-10-17
Conditionally rendered fields break onInput revalidation


halljus:
### Describe the bug and the expected behavior

Conditionally rendered Conform fields seem to lose their onInput revalidation once the condition changes. For example, for two (mutually exclusive) conditionally rendered inputs A and B:

1. Render InputA
2. Submit (firing the first validation)
3. Render InputB instead of InputA
4. Render InputA instead of InputB
5. InputA's onInput revalidation is now broken (you can submit again to fix it)

### Conform version

v0.9.0

### Steps to Reproduce the Bug or Issue

https://stackblitz.com/edit/remix-run-remix-dtzhfe?file=app%2Froutes%2F_index.tsx

Instructions are baked into the reproduction.

### What browsers are you seeing the problem on?

Chrome, Microsoft Edge

### Additional context

Note: our real schema is more complicated than this, with nested conditional fields, so things like discriminated unions didn't work as Zod doesn't support nested DUs, etc. so this `superRefine` approach was what worked best for us. But I doubt the schema format has anything to do with this anyway. To solve the revalidation problem, we had to hide/show fields via the `hidden` attribute rather than conditionally rendering them. It's a workaround, but probably not very idiomatic with regards to React, so we'd prefer to be able to conditionally render them instead if possible.

edmundhung:
If I understand correctly, you are expecting input A to be revalidated when you switch the type back.

The behaviour here is indeed working as per the current design. As the input has been unmounted whenever the type is changed, it is considered a new input that are never validated and follows `shouldValidate` option (which defaults to `onSubmit` in your example). This design is mainly inspired by how field list is handled. (i.e. If an item is deleted and then re-inserted, it will be considered a new item so it won't be revalidated right away). 

Does this design make sense to you? I find this question interesting as I am working on something that actually have the behaviour you want accidentally and I am now wondering what should be the correct behaviour, or at least the default behaviour here.

halljus:
Thanks, your explanation makes sense. It sounds like we'll just have to choose between the available behaviors based on the combination of conditional rendering + `shouldValidate` + `shouldRevalidate`.

[...]

PR #448, opened 2024-02-12
feat(conform-react): unstable useControl hook

edmundhung:
As reported on #306, #401, #443, the auto insert input feature on `useInputControl` is found to be problematic in several situations where you need to dynamically render the input. The current design is also not flexible in handling these cases. Although we can consider adding extra options to `useInputControl` for handling each cases, I believe it would be better stop supporting this and have developers deciding how they wanna render the hidden input themselves similar to the `useInputEvent` hook before v1.

As this would be a breaking change if we remove the feature directly from `useInputControl`, the react package will now export an additional `useControl` hook that works exactly the same as `useInputControl` except it will never insert an hidden input for you.

This is currently named `unstable_useControl` just so we can get some more feedbacks.

### DIFF
--- a/packages/conform-react/helpers.ts
+++ b/packages/conform-react/helpers.ts
@@ -1,10 +1,4 @@
-import type {
-	FormMetadata,
-	FieldMetadata,
-	Metadata,
-	Pretty,
-	Primitive,
-} from './context';
+import type { FormMetadata, FieldMetadata, Metadata, Pretty } from './context';
 
 type FormControlProps = {
 	key: string | undefined;
@@ -246,7 +240,7 @@ export function getFormControlProps<Schema>(
  * <input {...getInputProps(metadata, { type: 'radio', value: false })} />
  * ```
  */
-export function getInputProps<Schema extends Primitive | File[]>(
+export function getInputProps<Schema>(
 	metadata: FieldMetadata<Schema, any, any>,
 	options: InputOptions,
 ): InputProps {
@@ -292,12 +286,7 @@ export function getInputProps<Schema extends Primitive | File[]>(
  * <select {...getSelectProps(metadata, { value: false })} />
  * ```
  */
-export function getSelectProps<
-	Schema extends
-		| Exclude<Primitive, File>
-		| Array<Exclude<Primitive, File>>
-		| undefined,
->(
+export function getSelectProps<Schema>(
 	metadata: FieldMetadata<Schema, any, any>,
 	options: SelectOptions = {},
 ): SelectProps {
@@ -308,8 +297,8 @@ export function getSelectProps<
 
 	if (typeof options.value === 'undefined' || options.value) {
 		props.defaultValue = Array.isArray(metadata.initialValue)
-			? metadata.initialValue.map((item) => `${item ?? ''}`)
-			: metadata.initialValue;
+			? metadata.initialValue.map((item: string | undefined) => `${item ?? ''}`)
+			: metadata.initialValue?.toString();
 	}
 
 	return simplify(props);
@@ -330,9 +319,7 @@ export function getSelectProps<
  * <textarea {...getTextareaProps(metadata, { value: false })} />
  * ```
  */
-export function getTextareaProps<
-	Schema extends Exclude<Primitive, File> | undefined,
->(
+export function getTextareaProps<Schema>(
 	metadata: FieldMetadata<Schema, any, any>,
 	options: TextareaOptions = {},
 ): TextareaProps {
@@ -343,7 +330,7 @@ export function getTextareaProps<
 	};
 
 	if (typeof options.value === 'undefined' || options.value) {
-		props.defaultValue = metadata.initialValue;
+		props.defaultValue = metadata.initialValue?.toString();
 	}
 
 	return simplify(props);
--- a/packages/conform-react/index.ts
+++ b/packages/conform-react/index.ts
@@ -14,7 +14,10 @@ export {
 	FormStateInput,
 } from './context';
 export { useForm, useFormMetadata, useField } from './hooks';
-export { useInputControl } from './integrations';
+export {
+	useInputControl,
+	useControl as unstable_useControl,
+} from './integrations';
 export {
 	getFormProps,
 	getFieldsetProps,
--- a/packages/conform-react/index.ts
+++ b/packages/conform-react/index.ts
@@ -15,6 +15,7 @@ export {
 } from './context';
 export { useForm, useFormMetadata, useField } from './hooks';
 export {
+	Control as unstable_Control,
 	useControl as unstable_useControl,
 	useInputControl,
 } from './integrations';
--- a/packages/conform-react/index.ts
+++ b/packages/conform-react/index.ts
@@ -15,8 +15,8 @@ export {
 } from './context';
 export { useForm, useFormMetadata, useField } from './hooks';
 export {
+	useInputEvent as unstable_useInputEvent,
 	useInputControl,
-	useControl as unstable_useControl,
 } from './integrations';
 export {
 	getFormProps,
--- a/packages/conform-react/index.ts
+++ b/packages/conform-react/index.ts
@@ -15,7 +15,7 @@ export {
 } from './context';
 export { useForm, useFormMetadata, useField } from './hooks';
 export {
-	useInputEvent as unstable_useInputEvent,
+	useControl as unstable_useControl,
 	useInputControl,
 } from './integrations';
 export {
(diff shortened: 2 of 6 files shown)


edmundhung:
The new `useControl` API is still abstracting too much and does not work well in some scenarios. (e.g. checkbox group). So I am thinking about getting `useInputEvent()` back and focus only on dispatching events:

```tsx
import { useInputEvent } from '@conform-to/react';

function ExampleComponent() {
  // useInputEvent does not accept any options
  const control = useInputEvent();

  return (
     <div>
         {/* We will use the register callback ref to tell which elements the event will be dispatched from */}
         <input className="sr-only" name=".." ref={control.register} />
         <CustomInput
            onChange={event => control.change(event.target.value)}
            onFocus={contorl.focus}
            onBlur={control.blur}
         />
     </div>
  );
}
```

If you want to reuse the element from the custom input but have no access to the inner input, you can query it manually similar to v0:
```tsx
import { useInputEvent } from '@conform-to/react';

function ExampleComponent() {
  const control = useInputEvent();

  return (
     <div
       ref={element => {
          control.register(element?.closest('input'))
       }
     >
         <CustomInput
            name="..."
            onChange={event => control.change(event.target.value)}
            onFocus={contorl.focus}
            onBlur={control.blur}
         />
     </div>
  );
}
```

One downside is that you will need to manage the value yourself and make sure you passed the `key` to the custom input if you want to support reset / update form control:
```tsx
import { useInputEvent } from '@conform-to/react';

function ExampleComponent() {
  const [meta] = useField(name);
  const control = useInputEvent();

  return (
     <div>
         <input
            className="sr-only"
            key={meta.key}
            name={meta.name}
            ref={control.register}
         />
         <CustomInput
            key={meta.key}
            defaultValue={meta.initialValue}
            onChange={event => control.change(event.target.value)}
            onFocus={contorl.focus}
            onBlur={control.blur}
         />
     </div>
  );
}
```

Or, we can export another hook that manage reset the value for you:
```tsx
import { useInputEvent, useFieldValue } from '@conform-to/react';

function ExampleComponent() {
  const [meta] = useField(name);
  const [value, setValue] = useFieldValue(meta)
  const control = useInputEvent();

  return (
     <div>
         <input
            className="sr-only"
            name={meta.name}
            value={value}
            ref={control.register}
         />
         <CustomInput
            value={value}
            onChange={event => {
               setValue(event.target.value);
               control.change(event.target.value)
            }}
            onFocus={contorl.focus}
            onBlur={control.blur}
         />
     </div>
  );
}
```

Both `useFieldValue` and `useInputEvent` are used internally by the `useInputControl` hook but I couldn't figure out a good enough solution to serve both functionality in one 😅 

What do you think? @halljus and @EmilEinarsen

EmilEinarsen:
Here is my two cents

Having to supply the key prop to the custom input feels right and consistent with the library's approach to me. 
This way you could simply use `getInputProps` and not have to be concerned with it, right? 
```tsx
<input
      className="sr-only"
      {...getInputProps(meta)}
      ref={control.register}
  />
````

Comparatively, exposing `useFieldValue` just to reset/update the custom input, feels like a step backwards, leaning to heavy into a controlled approach.

Some other thoughts:
- I really dislike that the `useFieldValue` example results in three hooks being necessary and two setters being called on change.
- I often get errors in Chrome when I control the value of an input without supplying an onChange handler; so the control approach would need to be more verbose with an additional `onChange`.
- When registering the closest input, how would it be updated/reset?
- Wouldn't it be possible to reset the input through the reference provided by `control.register`?

**Comment by edmundhung** [comment]
Thanks for the quick resposne @EmilEinarsen! Really appreciate it.

> This way you could simply use getInputProps and not have to be concerned with it, right?

Yes. But you can use `defaultValue` even with `useInputControl` too as it will always try to update the input value for you. In case you wonder why: this is because controlled input makes it very common to have `onChange` called before its value state being updated and Conform will only see FormData with the previous value.

> Comparatively, exposing useFieldValue just to reset/update the custom input, feels like a step backwards, leaning to heavy into a controlled approach.

Conform forces react to re-mount the input by using `key`. This is unfortunately sub-optimal as this process will completely destroy the input state. e.g. If you are trying to update the same form input as you type, the cursor position will be lost when re-mounted. The `useFieldValue()` hook resolves this by handling it as a controlled input and no element will be re-mounted.

> I really dislike that the useFieldValue example results in three hooks being necessary and two setters being called on change.

Three hooks is definitely too much. But I just realize I can combine both hooks together just like `useInputControl()`. Would this look better?

```tsx
import { useControl } from '@conform-to/react';

function ExampleComponent() {
  const [meta] = useField(name);
  const control = useControl(meta);

  return (
     <div>
         <input
            className="sr-only"
            ref={control.register}
            {...getInputProps(meta, { type: 'text' })}
         />
         <CustomInput
            value={control.value}
            onChange={event => control.change(event.target.value)}
            onFocus={contorl.focus}
            onBlur={control.blur}
         />
     </div>
  );
}
```

> I often get errors in Chrome when I control the value of an input without supplying an onChange handler; so the control approach would need to be more verbose with an additional onChange.

I guess that's React warning you about uncontrolled input, right? In this case, using defaultValue setup with getInputProps would be a solution. For sure we can also pass a no-op function to onChange 😅 

> When registering the closest input, how would it be updated/reset?

If we are doing this, we are relying on the CustomInput to mange the value. That's why you will need to pass both `key` and `defaultValue` to the CustomInput as well.

> Wouldn't it be possible to reset the input through the reference provided by control.register?

We will need to reset the CustomInput as well. To do this, we either need to make it a controlled input with the value updated. or use the key to destroy and reinitialized it with the updated initialValue.

EmilEinarsen:
This looks great! This approach feels intuitive to me. 
It accomplish what I would expect, without adding needless complexity.

---

I'm looking forward to playing around with this! 
I have multiple forms with varied complexity of conditionally rendered/visible fields. Having the form and each field sync correctly was a hassle with `useInputControl`.

Thanks for your commitment to making this library so great!

edmundhung:
Awesome! I will have it released later today with an unstable prefix. Feel free to DM me if you find any issues using it.

halljus:
Sorry I wasn't able to reply until now, thanks for moving this forward! At a glance, I think I agree with Emil.

> If we are doing this, we are relying on the CustomInput to mange the value. That's why you will need to pass both `key` and `defaultValue` to the CustomInput as well.
> 
> We will need to reset the CustomInput as well. To do this, we either need to make it a controlled input with the value updated. or use the key to destroy and reinitialized it with the updated initialValue.

So, just to be clear - to include form reset functionality, would it look like this?

```tsx
import { useControl } from '@conform-to/react';

function ExampleComponent() {
  const [meta] = useField(name);
  const control = useControl(meta);

  return (
     <div>
         <input
            className="sr-only"
            ref={control.register}
            {...getInputProps(meta, { type: 'text' })}
         />
         <CustomInput
            key={meta.key} // for reset? or not needed?
            defaultValue={meta.initialValue} // for reset? or not needed?
            value={control.value}
            onChange={event => control.change(event.target.value)}
            onFocus={contorl.focus}
            onBlur={control.blur}
         />
     </div>
  );
}
```

Or are `key` and `defaultValue` unnecessary because we're controlling it by setting `value`? Will a form reset automatically reset a useControl value?

edmundhung:
You need `key` only if you are using the uncontrolled approach (i.e. `defaultValue`). With `useControl`, you will only need the `value` as we are making it controlled and the value will be updated automatically with update / reset intent.

edmundhung:
If you find it confusing, maybe it would be easier to just read the [code](https://github.com/edmundhung/conform/pull/448/files#diff-1cf5d16fe07a3a8e7f2154f192d4f2181fc51e1c5ed40a9fe3c5b3246543f73eR268-R287) and see how we reset the value based on `key`

EmilEinarsen:
> Awesome! I will have it released later today with an unstable prefix. Feel free to DM me if you find any issues using it.

Fantastic! 
Where would I DM you; does GitHub have a DM feature I'm not aware of, or Twitter?

edmundhung:
> > Awesome! I will have it released later today with an unstable prefix. Feel free to DM me if you find any issues using it.
> 
> Fantastic! Where would I DM you; does GitHub have a DM feature I'm not aware of, or Twitter?

You can find me on both Twitter or Discord :) Twitter handle on profile / GitHub username on discord

edmundhung:
I am just wrapping this up and decided to export a render props version as well. (i.e. `<Control />`) It will be useful in case of multiple input per field (e.g. [CheckboxGroup](https://github.com/edmundhung/conform/pull/448/files#diff-6efa6b92dc895910895a26605f166103bed6a56f18c49617eef0ca3b5d3a87f6R412-R464)) or if you prefer inlining without creating an additional component.

lilouartz:
When is `unstable_useControl` going to become stable?

PR #449, opened 2024-02-13
docs: Update tutorial.md

ngbrown:
Fix use of sendMessage in `action()`. `submission.value` should be used instead of `result.data` and `action()` should end with `redirect()` like initial example.


edmundhung (approved):
Thanks

PR #465, opened 2024-02-18
feat: add object and array fields to getYupConstraint validation

gglee89:
### What was done:
- implementation of the `getYupConstraint` utility function.
- Added validation for:
    - `timestamp`
    - `array`
    - `nested object`
    - `list`
    - `tuple`
    	- ⚠️ until yup is updated to `v1` we added a custom method for tuples which outputs tuples as:
    		- `tuple.0`:  {...}
    		- `tuple.1`: {...}


packages/conform-dom/form.ts
```diff
@@ -105,8 +105,8 @@ export type Constraint = {
 	required?: boolean;
 	minLength?: number;
 	maxLength?: number;
-	min?: string | number;
-	max?: string | number;
+	min?: string | number | Date;
+	max?: string | number | Date;
```
gglee89, line 109: not sure about this part. If it's safe to extend the type with `Date`. This was added to allow the developer to set `min` and `max` Dates for the timestamp field.

tests/conform-yup.spec.ts
```diff
@@ -1,41 +1,65 @@
...
 
+// Adding custom method tuple until Yup is updated to v1 where the tuple method is built-in
+// ref: https://github.com/jquense/yup/issues/528#issuecomment-916885944
+yup.addMethod(yup.array, 'tuple', function (schema) {
+	if (!this.isType(schema)) yup.ValidationError();
+	return yup.object({
+		...Object.fromEntries(Object.entries(schema)),
+	});
+});
+
```
edmundhung, line 15: It might be better to ignore tuple if it is not built-in since we can't ensure that the user has the same implementation / understanding of `tuple`.

tests/conform-yup.spec.ts
```diff
@@ -1,41 +1,65 @@
...
+				.array(
+					yup.string().required('required').oneOf(['a', 'b', 'c'], 'invalid'),
+				)
+				.required('required')
 				.min(3, 'min'),
 			nested: yup
 				.object({
 					key: yup.string().required('required'),
 				})
-				.test('nested', 'error', () => false),
```
edmundhung, line 29: Can you keep this check? I wanna make sure group level fields are still validated properly.

tests/conform-yup.spec.ts
```diff
@@ -79,11 +104,48 @@ describe('conform-yup', () => {
...
+			tuple: {
+				required: true,
+			},
+			'tuple.0': {
+				required: true,
+				minLength: 3,
+			},
+			'tuple.1': {
+				max: 100,
+			},
```
edmundhung, line 145: I am not sure how tuple works. There is no dedicated logic for it. Does it depends on your implementation of the `tuple` method above? Also, tuple has to be in the format `tuple[0]` / `tuple[1]`. 

It might be better leaving tuple unsupported 🤔

edmundhung:
Sorry for the late review 🙏🏼 @gglee89

tests/conform-yup.spec.ts
gglee89: agree. 👍  Thanks for the feedback. Tuple will be natively supported once `yup` version is updated to `v1`, so you can use that https://github.com/jquense/yup/issues/528#issuecomment-916885944

gglee89:
Thanks for the feedback @edmundhung 
I've made the changes to leave `tuple` unsupported. Makes sense, especially because it's natively supported on on `yup v1`.

edmundhung (dismissed):
This is awesome. Thanks!

edmundhung (approved):

PR #477, opened 2024-02-21
Fix codesample typo in tutorial.md

nphmuller:
Change \<form\> to \<Form\> since the Remix Form component is used in the tutorial.

This already was in place for all code samples except for one instance.


nphmuller:
And thank you for this awesome library! Really looking forward to using it in the coming weeks.

PR #483, opened 2024-02-23
docs: fix change lastSubmission to lastResult in validation.md

ozanbulut:
Update the example in validation.md
Change `lastSubmission` to `lastResult`


edmundhung (approved):
Thanks!

PR #498, opened 2024-02-27
docs: Correct "field." to "fields." and "validate" to "validates"

hawkcookie:
This PR corrects some typo.

I have used since version 0, it became even more user-friendly with version 1. Thank you for your library.


edmundhung (approved):
Thank you!

PR #508, opened 2024-03-03
Fix `getFieldset` with interfaces and nullable schemas

### LINKED ISSUE #499: Property 'getFieldset' does not exist on type 'FieldMetadata'

I have a simple form:
```ts
    lastResult,
    defaultValue: user,
    shouldValidate: "onBlur",
    onValidate({ formData }) {
      return parseWithZod(formData, { schema: dataSchema });
    },
  });
  const profile = fields.profile.getFieldset();
```
The LSP and `lint` does not report an error. However, `typecheck` is:
```
error TS2339: Property 'getFieldset' does not exist on type 'FieldMetadata<JsonifyObject<{ createdAt: Date; userId: string; ... }> | null, JsonifyObject<...>, string[]>'.
```

### Conform version

1.0.2

Run `typecheck` while using `getFieldset`

_No response_

### Screenshots or Videos

_No response_

### Additional context

_No response_

aaronadamsCA:
This should fix #499, specifically the types failing to reveal `getFieldset` in these two scenarios:

- Your `FieldSchema` is an interface, not a type.
- Your `defaultValue` is nullable.

I'm filing this as a draft, as I have no idea what kinds of tests to write for this.

Known existing issues this PR doesn't try to fix:

- `getFieldset` is incorrectly available on other object types, e.g. `Date`.

### DIFF
--- a/packages/conform-react/context.tsx
+++ b/packages/conform-react/context.tsx
@@ -83,11 +83,11 @@ type SubfieldMetadata<
 	FormSchema extends Record<string, any>,
 	FormError,
 	CombinedSchema = Combine<Schema>,
-> = Exclude<Schema, undefined> extends Array<infer Item>
+> = NonNullable<Schema> extends Array<infer Item>
 	? {
 			getFieldList: () => Array<FieldMetadata<Item, FormSchema, FormError>>;
 	  }
-	: Exclude<Schema, undefined> extends Record<string, unknown>
+	: NonNullable<Schema> extends object
 	? {
 			getFieldset: () => Required<{
 				[Key in keyof CombinedSchema]: FieldMetadata<


aaronadamsCA:
Ah, okay, I see now that these are tested via a playground package type check, so now I know what to run locally to test changes.

I fixed a weird problem caused by TypeScript quirks:

- `NonNullable` is implemented as `T & {}`, since `{}` means "any non-nullish value".
  - Therefore `NonNullable<unknown>` resolves to `{}`
- But `{} extends object` resolves to `true`, probably for legacy reasons.
  - Therefore `NonNullable<unknown> extends object` resolves to `true` (yikes!).

I also updated the tests to use `interface` instead of `type`. They fail without this change and pass with it.

edmundhung (approved):
Thank you!

PR #510, opened 2024-03-06
refactor(react): handle input, blur, reset by react event

QzCurious:
Although no test failed, this change could potentially change the convention of how conform-react is used. It is now requiring users to use getFormProps exclusively. Because it no longer handle those events implicitly.


QzCurious:
close #476

PR #517, opened 2024-03-14
docs: Fix missing map index in intent-button.md

hawkcookie:
I fixed that the index was not written in intent-button.md.


edmundhung (approved):
Thanks!

PR #541, opened 2024-03-25
Avoid type widening with enum-typed values in FormValue

tmcw:
The current generic type for `FormValue` is this:

```ts
export type FormValue<Schema> = Schema extends
	| string
	| number
	| boolean
	| Date
	| bigint
	| null
	| undefined
	? string | undefined
```

In this, if you have an enum or constant string type, then it will get widened into a `string` type. Turning `Date` and `null` types into string | undefined makes sense, but it is kind of just inconvenient to lose type information if you have an enum.

This tweaks the generic type so that if you have a type that `extends string`, then the `FormValue` generic uses that type instead of widening to string.

### DIFF
--- a/packages/conform-dom/form.ts
+++ b/packages/conform-dom/form.ts
@@ -58,14 +58,9 @@ export type DefaultValue<Schema> = Schema extends
 	? { [Key in keyof Schema]?: DefaultValue<Schema[Key]> } | null | undefined
 	: string | null | undefined;
 
-export type FormValue<Schema> = Schema extends
-	| string
-	| number
-	| boolean
-	| Date
-	| bigint
-	| null
-	| undefined
+export type FormValue<Schema> = Schema extends string
+	? Schema | undefined
+	: Schema extends number | boolean | Date | bigint | null | undefined
 	? string | undefined
 	: Schema extends File
 	? File | undefined


edmundhung:
This is tricky... because conform does not validate the form value. The type you got is more or less a faith that the type should be __mostly correct__ if you are using Conform properly. But it couldn't stop you from put a hidden input with an invalid enum which will be reflected on form value regardless. 😅 

Maybe it is alright to made the same assumption with enum here... What do you think?

jansedlon:
I think that hat's what typescript's about, constraining things to be something we want even though Javascript allows to do about anything. If you use typescript incorrectly or if you make a mistake and mistakenly widen a type to things you don't want to, you will fuck yourself up anyway :D 

On the other hand since inputs are strings / files by design.. I would slightly incline to have it widened and put the responsibility on the developer. I know, I also don't like not having the convenience but.. It forces the developer to put more checks in place.

lifeiscontent:
might be worth using the string interpolation type:

https://www.typescriptlang.org/play?#code/C4TwDgpgBAYg9gJwLYDUCGAbArhAPAZQGMALCJNAPigF4ojTyoIAPYCAOwBMBnAKCigAfKN2AIAluwDm-IVHZYkAIwgJZwpeKmTg6qErhwMENOz0KMGPVi4QAZpIidZAfigADACQBvemTQAvu5yNpz2js4CAFx0JP5MrBw8sOLGrinGIbYO7E6yMbkAbqoA3Ly8oJAZEOjY0LTwyLU4uDCpEBRlldD4YpJSzfWwiKiYLaIS0p0V4D190gCq7OJw7IM0w01jeABEAIw7cjsATDvT3VAAcooqCOsNI4O4Csqq57NXN6pLK2vbG41RnVcHs5MdprwAPSQqA2bhYBDQcR2KDAYjibhQADucCwGE4+mgLEghDYnAANPosMAoCBcVApBAaWjoNpRAA6GZVABCWgAkuxgPdNkCWpptIL3jz+YKfqthYCnnt2GD2FLoEswjknArHttcKFwrlOOrPpZdVtgRYMNMgA

PR #544, opened 2024-03-26
feat(guide): added language switcher dropdown and Japanese language support.

coji:
This pull request adds language switcher dropdown and Japanese language support to the Conform documentation site.

CleanShot 2024-03-26 at 15 47 40@2x

Key changes:
- Modified layout.tsx to load localized navigation menu and added language switcher dropdown
- Added allLanguages constant to define supported languages and their properties
Updated getMetadata, getFileContent, and getDocPath functions to handle language-specific logic
- Updated route files to use getDocPath for language-specific document paths

These changes enable a localized experience for Japanese users within the guide submodule, with the ability to switch between English and Japanese using the language dropdown in the header.

### DIFF
--- a/guide/app/layout.tsx
+++ b/guide/app/layout.tsx
@@ -6,76 +6,142 @@ import {
 	usePageLoaderData,
 	MainNavigation,
 } from '~/components';
-import { logo } from '~/util';
+import { allLanguages, getLanguage, logo } from '~/util';
 
-const menus: Menu[] = [
-	{
-		title: 'Getting Started',
-		links: [
-			{ title: 'Overview', to: '/' },
-			{ title: 'Tutorial', to: '/tutorial' },
-			{ title: 'Upgrading to v1', to: '/upgrading-v1' },
-			{ title: 'GitHub', to: 'https://github.com/edmundhung/conform' },
-		],
-	},
-	{
-		title: 'Guides',
-		links: [
-			{ title: 'Validation', to: '/validation' },
-			{ title: 'Nested object and Array', to: '/complex-structures' },
-			{ title: 'Intent button', to: '/intent-button' },
-			{ title: 'Accessibility', to: '/accessibility' },
-			{ title: 'Checkbox and Radio Group', to: '/checkbox-and-radio-group' },
-			{ title: 'File Upload', to: '/file-upload' },
-		],
-	},
-	{
-		title: 'Integration',
-		links: [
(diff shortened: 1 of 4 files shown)


guide/app/layout.tsx
```diff
@@ -106,16 +172,30 @@ export function Guide({
...
 						to="/"
 					>
 						{logo}
 					</Link>
+					<select
+						className="bg-zinc-900 text-xs rounded-md py-1"
+						defaultValue={language.code}
+						onChange={(e) => {
+							const selectedLanguage = getLanguage(e.target.value);
+							window.location.href = `//${selectedLanguage.domain}${location.pathname}${location.search}`;
```
lifeiscontent, line 188: should use `navigate` from useNavigate hook

guide/app/layout.tsx
```diff
@@ -106,16 +172,30 @@ export function Guide({
...
 						title="Conform"
 						to="/"
 					>
 						{logo}
 					</Link>
+					<select
+						className="bg-zinc-900 text-xs rounded-md py-1"
+						defaultValue={language.code}
+						onChange={(e) => {
+							const selectedLanguage = getLanguage(e.target.value);
```
lifeiscontent, line 187: does the same thing, but... semantics
```suggestion
							const selectedLanguage = getLanguage(e.currentTarget.value);
```

guide/app/layout.tsx
coji: This place does not work unless window.location.href is used.

Transitions due to language switching are external URLs. The useNavigate/navigate of react-router supports external URLs and navigate does not allow page transitions.
The react-router author is told to use window.location.href in [this issue](
https://github.com/remix-run/react-router/issues/8317#issuecomment-974577787).

guide/app/layout.tsx
coji: Fixed to currentTarget. Thank you.

guide/app/layout.tsx
lifeiscontent: @coji ah, sorry, I didn't look close enough, yes you are right!

edmundhung (approved):
Thanks @coji! I just made the logo smaller on desktop so the language switcher can fit in better on desktop, but everything else looks good to me! 

Personally I would also prefer the language switcher to work before JS, but that might requires changes on the layout so I would leave it as a future improvement.

PR #548, opened 2024-03-28
Feat/update at index

jansedlon:
Hey @edmundhung. I am in need to update a list item at a specific index so I tried to implement it. I welcome any suggestions. There's no contribution guide so I tried to scrap the code and came up with this :D

Use case:
I have a list of attachments in a community post. When a user wants to upload a file, I want to immediately insert it into a list with a loading state and when the upload is finished, I want to update the url and state at a specified index.

For now I have done it using remove and insert which works but is suboptimal.

### DIFF
--- a/packages/conform-dom/form.ts
+++ b/packages/conform-dom/form.ts
@@ -376,7 +376,8 @@ function handleIntent<Error>(
 		}
 		case 'insert':
 		case 'remove':
-		case 'reorder': {
+		case 'reorder':
+		case 'updateAt': {
 			if (initialized) {
 				meta.initialValue = clone(meta.initialValue);
 				meta.key = clone(meta.key);
@@ -1012,6 +1013,7 @@ export function createFormContext<
 		reset: createFormControl('reset'),
 		update: createFormControl('update'),
 		insert: createFormControl('insert'),
+		updateAt: createFormControl('updateAt'),
 		remove: createFormControl('remove'),
 		reorder: createFormControl('reorder'),
 		subscribe,
--- a/packages/conform-dom/submission.ts
+++ b/packages/conform-dom/submission.ts
@@ -198,7 +198,8 @@ export function parse<FormValue, FormError>(
 				break;
 			}
 			case 'remove':
-			case 'reorder': {
+			case 'reorder':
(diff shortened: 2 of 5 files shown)


jansedlon:
Hey @edmundhung. I am in need to update list item at a specific index so I tried to implement it. I welcome any suggestions. There's no contribution guide so I tried to scrap the code and came up with this :D 

Use case:
I have a list of attachments in a community post. When a user wants to upload a file, I want to immediately insert it into a list with a loading state and when the upload is finished, I want to update the url and state at a specified index.

For now I have done it using remove and insert which works but is suboptimal.

edmundhung:
I haven't look into the detail yet. But the update intent should work anywhere, you can point it to the list and update the whole list or point it to a specific index and update the value just like any other field.

jansedlon:
From what I saw, it accepts a list of values. If I try to update a specific index, the field doesn't exist yet.

Example
```
form.insert({
  name: fields.attachments.name,
  value: {
    // ...value
  }
});

// code

fetch().then(() => {
  // I want to update here but the reference to the fieldlist does not have the inserted field
})
```

jansedlon:
This works, but it's meh :D 
```
form.update({
          name: `attachments[${someDeterminedIndex}]`,
          value: {
            // data
          },
        });
```

edmundhung:
How about making `form.update` to accept an optional index? Then you could do this:
```tsx
form.update({
  name: fields.attachments.name,
  index: someDeterminedIndex,
  value: {
    // data
  },
});
```

jansedlon:
Absolutely, that could be the case. However the update logic is much more complex and I'm not sure I'd do it correctly

edmundhung:
Here you go: #555

jansedlon:
Hey, thank you 🙏🙏 Sorry, haven't had time to try out, will tomorrow

edmundhung:
It's all good :) You can try it out on [v1.1.0-pre.0](https://github.com/edmundhung/conform/releases/tag/v1.1.0-pre.0) now.

jansedlon:
@edmundhung Works flawlessly 😍

PR #556, opened 2024-04-01
rebase to latest main

coji:
rebase


coji:
@edmundhung I was hoping to be able to do this merge myself, but I don't seem to have the authority to do so. What should I do?

CleanShot 2024-04-01 at 20 35 26@2x

edmundhung:
@coji As we now have the logic to fetch markdown file based on langauge, I think it would be simpler to just have the remix app figuring out the language based on the hostname. I am just adding this logic right now and I will merge the `ja` branch back to main if you are fine with this :)

edmundhung:
Done. We are now serving the Japanese docs from the main branch as well.

coji:
@edmundhung 
Thanks for taking it in. I'm glad you did!

PR #566, opened 2024-04-04
chore: upgrade playground

lifeiscontent:
@edmundhung I've upgrade the playground and I've run into a couple of issues running playwright, I imagine if we figure out what's going on, we might be able to track down some of the issues we've been discussing.

if you think this is worthwhile doing, I'll continue forward.

here's a few line items that are still needed to be done.

1. add lint-staged back
  1. had to remove due to ESLint and how it was setup (we should discuss)
2. update remix guide app to V2 (can work on this next)
3. update all example apps to make sure they're in working order
4. potentially restructure the packages so they all contain their own eslint config. so we can fix 1.

let me know how you'd like to proceed.

if you checkout this branch and run the following:

```
pnpm i
pnpm playground build
pnpm playground test:e2e
```

you can see a lot of the tests failing (no package code was changed)


edmundhung:
Thanks for working on this! I would be happy to review PRs for upgrading the dependencies and improving the setup. Just that it would be really helpful if you can minimize the amount of changes in each PR so I would feel more confident merging them 🙏🏼

> add lint-staged back

What was the issue with lint-staged? It looks like your have it removed

> had to remove due to ESLint and how it was setup (we should discuss)

That would be awesome. I did run into some eslint issue recently which I have no time to investigate further 😅 

> update remix guide app to V2 (can work on this next)

You can leave this to me. I have some works done for this already.

> update all example apps to make sure they're in working order

What do you mean?

> potentially restructure the packages so they all contain their own eslint config. so we can fix 1.

I would love to learn more. Thanks!

edmundhung:
Also, I have checked out your branch and all tests are passing for me still 😅  It might be some OS issue as you pointed out before (I still haven't fix those..)

lifeiscontent:
@edmundhung I've updated everything, now all we need to do is figure out the lint-staged stuff. 

just a rough outline:

1. I've basically made each package operate like its own repository.
2. I've added a few helper commands to operate within the guide/playground more easily from the root

some notes:

for lint-staged / eslint stuff

we can add a `pre-commit` script to each package that wants to take advantage of it, and do the operations there.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+	"eslint.workingDirectories": [{ "mode": "auto" }]
```
lifeiscontent, line 2: @edmundhung not sure if you use VSCode, but this tells the ESLint plugin to use each .eslintrc within its package.

CONTRIBUTING.md
```diff
@@ -0,0 +1,13 @@
...
+## Installation
+
+```bash
+pnpm i
+```
+
+## Development
+
+```bash
+pnpm dev
```
lifeiscontent, line 12: I still need to figure out how to get this to work properly, some ports need to be remapped, probably in the examples repo.

examples/chakra-ui/package.json
```diff
@@ -2,20 +2,20 @@
...
-		"react": "^18.2.0",
-		"react-dom": "^18.2.0",
+		"@chakra-ui/react": "2.8.2",
+		"@conform-to/react": "workspace:*",
+		"@emotion/react": "11.11.4",
+		"@emotion/styled": "11.11.5",
+		"framer-motion": "11.0.24",
+		"react": "18.2.0",
+		"react-dom": "18.2.0",
 		"react-scripts": "5.0.1"
```
lifeiscontent, line 12: I've pinned all deps to a specific version, and when we want to upgrade we can run `pnpm update -r --interactive --latest` and it'll swap out all the deps.

pinning versions is imo the best way to build software because you don't have random version bumps come into the project to screw things up

guide/worker-configuration.d.ts
```diff
@@ -0,0 +1,9 @@
+// Generated by Wrangler on Fri Apr 05 2024 00:43:41 GMT-0700 (Pacific Daylight Time)
+// by running `wrangler types`
+
+interface Env {
+	CACHE: KVNamespace;
+	CF_PAGES_BRANCH: "main";
+	LANGUAGE: "en";
```
lifeiscontent, line 7: @edmundhung I'm not sure how to get these two vars to work the way you had them

packages/conform-dom/tsconfig.json
```diff
@@ -1,15 +1,22 @@
...
-		"target": "ES2022",
-		"module": "ES2022",
+		"lib": ["DOM", "DOM.Iterable", "ES2022"],
+		"types": ["vite/client"],
+		"isolatedModules": true,
+		"esModuleInterop": true,
+		"module": "ESNext",
 		"moduleResolution": "Bundler",
-		"allowSyntheticDefaultImports": true,
-		"noUncheckedIndexedAccess": true,
```
lifeiscontent, line 8: should probably add these options back

packages/conform-react/tsconfig.json
```diff
@@ -1,16 +1,23 @@
...
-		"target": "ES2022",
-		"module": "ES2022",
-		"moduleResolution": "Bundler",
+		"lib": ["DOM", "DOM.Iterable", "ES2022"],
+		"types": ["vite/client"],
+		"isolatedModules": true,
+		"esModuleInterop": true,
 		"jsx": "react-jsx",
-		"allowSyntheticDefaultImports": true,
-		"noUncheckedIndexedAccess": true,
```
lifeiscontent, line 9: should probably add this back

packages/conform-yup/tsconfig.json
```diff
@@ -1,15 +1,22 @@
...
-		"target": "ES2022",
-		"module": "ES2022",
+		"lib": ["DOM", "DOM.Iterable", "ES2022"],
+		"types": ["vite/client"],
+		"isolatedModules": true,
+		"esModuleInterop": true,
+		"module": "ESNext",
 		"moduleResolution": "Bundler",
-		"allowSyntheticDefaultImports": true,
-		"noUncheckedIndexedAccess": true,
```
lifeiscontent, line 8: should add this back

packages/conform-zod/tsconfig.json
```diff
@@ -1,15 +1,23 @@
 {
-	"compilerOptions": {
-		"lib": ["ES2022", "DOM", "DOM.Iterable"],
-		"target": "ES2022",
-		"module": "ES2022",
-		"moduleResolution": "Bundler",
-		"allowSyntheticDefaultImports": true,
-		"noUncheckedIndexedAccess": true,
```
lifeiscontent, line 8: should add this back

lifeiscontent:
@edmundhung do you like where this PR is headed?

1. upgraded all packages and apps to latest version (except for yup)
2. migrated code to latest standards (LoaderFunctionArgs/etc)
3. decoupled build system into individual packages (via vite)
4. deduped packages across repo by using pinned versions
5. got ESLint working across packages

TBD:

7. get pre-commit back
8. need to add an eslint file for conform related packages

edmundhung:
I like a lot of the changes you have made, thank you! But I really need you to split some of these changes to a smaller PRs:

1. Prettier upgrade and format run
2. Updating the remix example and playground app to v2 (This could be 2 PR as well)
3. Updating the rest of the packages and pin the dependency version
4. Fix eslint setup (if you figure out the solution already)

Some of the changes are also unclear to me why it is needed:
- What do we gain from building the packages with vite? I would like to learn more your rationale first. This change must also be on its own PR. 
- Why moving the build / test setup to individual packages.

lifeiscontent:
#### What do we gain from building the packages with vite? I would like to learn more your rationale first. This change must also be on its own PR.

All great questions, we actually don't need to move to vite, it was just something I did quickly so I didn't have to migrate the rollup configs. the main factor of why moving each package to its own build process is as follows:

first, using `prepare` eliminates the need for a build step entirely as outlined in the npm docs (this practice has existed even before CoffeeScript was a thing)

when you run `pnpm i` the prepare step will take care of building it also runs before publishing.

https://docs.npmjs.com/cli/v10/configuring-npm/package-json#devdependencies
use lib directory for transpiled JS as defined by the package.json spec: https://wiki.commonjs.org/wiki/Packages/1.0#Package_Directory_Layout

#### Why moving the build / test setup to individual packages.

my rational here is because within a monorepo the root package should have the minimal amount of deps for operating the monorepo, I figured moving the tests to playground made a lot of sense because its effectively the app you use for testing anyway, this allows us to keep the list of workspace dependencies to a minimum rather than having them hang around in all other packages (which is what happens in the workspace package)

edmundhung:
Personally I would prefer not to change the build tool for now. Although Vite use rollup as well, there might be slight difference on the result built that might becomes a breaking change to some people. But I am happy to revisit this idea on v2.

We are running the playwrights test against the playground remix app at the moment. But I would love to have it running against each examples eventually so we could cover other integrations such vite + react-router / Remix SPA mode or even NextJS. That's why I left it at the root project. If you have any idea how to make this happens, I would love to learn more from you as well.

I will merge a few more PRs today and release v1.1.0. Once it is out, I am happy to merge your changes in the next couple days if you can submit smaller PRs as I suggested above. Thank you!

lifeiscontent:
@edmundhung I see, that makes sense, I'll see if I push up a few PRs over the weekend before I get started on that though, how would you like me to deal with the rollup config? here's a few ideas I have in my head.

1. create a `config` package for all shared configs
2. create an individual config for each package.
3. create not a rollup config, but rather a helper that becomes a rollup config in the root of the project which all packages import.

edmundhung:
I might go with 2. They will look totally the same for now but some packages might have specific setup in the future.

PR #584, opened 2024-04-15
Update helpers.ts

lifeiscontent:
the docs say we should pass `false` but because of the previous type def, this was not possible


edmundhung (approved):
Thanks!

PR #594, opened 2024-04-22
fix: defaultValue

lifeiscontent:
@edmundhung I would write a test for this, but it's pretty difficult for me to do, the fix here is as follows:

when serializing the form if a `getButtonProps` defaultValue has a value (key or direct value) that is `null` it should be preserved as its a value that is JSON Serializable and if a schema says it can be nullable, it should pass that check


lifeiscontent:
@edmundhung nvm, this PR doesn't fix the issue, I'm going to try modify the code locally on my end before submitting another PR for this, or if you have some ideas on how to fix this serialization issue, please let me know

PR #595, opened 2024-04-23
chore: simplify types

lifeiscontent:
simplifying types to resolved types.

1. `Array<T>` -> `T[]`
2. `Record<string | number | symbol, unknown>` -> `Record<keyof any, unknown>` as it resolves to the same thing

### DIFF
--- a/packages/conform-dom/formdata.ts
+++ b/packages/conform-dom/formdata.ts
@@ -30,25 +30,23 @@ export function getFormData(
  * const paths = getPaths('todos[0].content'); // ['todos', 0, 'content']
  * ```
  */
-export function getPaths(name: string | undefined): Array<string | number> {
+export function getPaths(name: string | undefined): (string | number)[] {
 	if (!name) {
 		return [];
 	}
 
-	return name
-		.split(/\.|(\[\d*\])/)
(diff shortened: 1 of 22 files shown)


packages/conform-dom/formdata.ts
```diff
@@ -148,9 +146,7 @@ export function getValue(target: unknown, name: string): unknown {
 /**
  * Check if the value is a plain object
  */
-export function isPlainObject(
-	obj: unknown,
-): obj is Record<string | number | symbol, unknown> {
+export function isPlainObject(obj: unknown): obj is Record<keyof any, unknown> {
```
lifeiscontent, line 149: @edmundhung also simplified object types to use `keyof any` which resolves to the same type.

edmundhung:
I am sorry but this is probably a perspective of preference here as I don't see this simplifying the types. 😅 

Personally I prefer `Array<T>` over `T[]` especially with complex type, e.g. `(A | B | C)[]` is much harder to read then `Array<A | B | C>`. Similarly, `Record<keyof any, unknown>` requires an additional thought of what `keyof any` is.

lifeiscontent:
@edmundhung hey no worries, I've closed the branch :)

PR #597, opened 2024-04-23
fix: getButtonProps types

### LINKED ISSUE #514: Insert defaultValue is not typed if field is a nullable or optional array

I am creating some dynamic fields in my form using the `form.insert` method, but noticed that I was getting a type error  whilst trying to add some default values. It would appear that when the field is an array and is optional/nullable then the `defaultValue` type becomes a string. Here is a simplified example

```ts
export const schema = z.object({
  test: z.array(z.object({ nested: z.string() })).nullable() // or .optional()
})
```

```ts
<button
  {...form.insert.getButtonProps({
    name: fields.test.name,
    defaultValue: { // <-- Type '{ nested: string; }' is not assignable to type 'string'.
      nested: "I am nested",
    },
  })}
>
  Click me
</button>
```

It's fairly easy to work aroud this situation by making the field required and setting the default value of the form to an empty array, however, I thought I would report in case this was not deliberate!

Maintaining the type inside of the array would make the most sense.

### Conform version

v1.0.3

1. Create a simple schema with an an optional/nullable array as a field.
2. Use the field metadata to insert values in the array.
3. See defaultValue is types as a string and not the type within the schema.

Chrome, Firefox, Safari

### Screenshots or Videos

_No response_

### Additional context

_No response_

lifeiscontent:
closes #514

### DIFF
--- a/packages/conform-dom/submission.ts
+++ b/packages/conform-dom/submission.ts
@@ -320,7 +320,7 @@ export type ResetIntent<Schema = any> = {
 		  };
 };
 
-export type UpdateIntent<Schema = unknown> = {
+export type UpdateIntent<Schema = any> = {
 	type: 'update';
 	payload:
 		| {
@@ -339,7 +339,7 @@ export type UpdateIntent<Schema = unknown> = {
 		  };
 };
 
-export type RemoveIntent<Schema extends Array<any> = any> = {
+export type RemoveIntent<Schema = any> = {
 	type: 'remove';
 	payload: {
 		name: FieldName<Schema>;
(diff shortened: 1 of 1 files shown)


packages/conform-dom/submission.ts
```diff
@@ -339,15 +339,15 @@ export type UpdateIntent<Schema = unknown> = {
 		  };
 };
 
-export type RemoveIntent<Schema extends Array<any> = any> = {
```
edmundhung, line 342: These checks are intentional to make sure the user passed a field with an array type since they won't work with non-array field. We might be able to fix it by `Schema extends Array<any> | null | undefined` instead

lifeiscontent: the suggestion you've proposed doesn't work btw.

edmundhung (approved):
Not sure how I broke the types for array but this is definitely an improvement over what we have. LGTM

PR #601, opened 2024-04-24
Remove usage of `instanceof` for Zod schemas

colinhacks:
I got a weird bug report on Twitter here: https://twitter.com/Varkoffs/status/1783118133848400070

It seems to be due to a weirdness in Remix, where both the ESM and CommonJS versions of Zod are loaded and used during server initialization. Somehow a schema declared in a Remix page (using ESM) was getting passed into a `parseWithZod` call that was using CJS internally to do `instanceof` checks. Because each copy of Zod has separate class definitions, the `instanceof` checks fail. 

This is by no means an issue with conform, but one possible solution is for conform to avoid `instanceof`. There are other scenarios where both builds of a package can be loaded simultaneously. Some are described by isaacs (founder of npm) here: https://github.com/isaacs/tshy?tab=readme-ov-file#dual-package-hazards 

The tldr is that its best to avoid `instanceof` when possible, especially for libraries that are referencing classes from their dependencies. 

This PR technically relies on "internal" `_def` APIs but these haven't changed in a long time and you can consider them stable. (Zod 4 will break this, but that's an issue for another day.) Apologies for some of the hackiness/casting but I didn't see a way to avoid it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,7 @@ node_modules/
 /test-results/
 /playwright-report/
 /playwright/.cache/
+
+# DS Store
+.DS_Store
+**/.DS_Store
--- a/packages/conform-zod/coercion.ts
+++ b/packages/conform-zod/coercion.ts
@@ -1,31 +1,25 @@
 import {
-	type ZodType,
-	type ZodTypeAny,
-	type output,
-	ZodString,
-	ZodEnum,
-	ZodLiteral,
-	ZodNumber,
-	ZodBoolean,
-	ZodDate,
 	ZodArray,
-	ZodBigInt,
-	ZodNativeEnum,
 	ZodObject,
-	ZodLazy,
 	ZodIntersection,
 	ZodUnion,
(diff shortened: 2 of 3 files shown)


Varkoff:
Thank you for offering a submission.

Codesandbox https://codesandbox.io/p/github/Varkoff/remix-monorepo-zod-issue/main?import=true
Repository https://github.com/Varkoff/remix-monorepo-zod-issue

edmundhung:
Thanks for making the changes, @colinhacks! 

> This PR technically relies on "internal" _def APIs but these haven't changed in a long time and you can consider them stable. (Zod 4 will break this, but that's an issue for another day.) Apologies for some of the hackiness/casting but I didn't see a way to avoid it.

I think this is a good workaround. I ran into this issue before but I didn't realize I could get rid of `instanceOf` completely. The existing logic relies on the internal APIs already so what you did is totally fine. 👍🏼

edmundhung (approved):

PR #602, opened 2024-04-25
chore: fix helpers

lifeiscontent:
- 1. Fix `getInputProps` to return a defaultValue if the defaultValue is a number
- 2. Fix `getInputProps` to return `defaultValue` instead of `value` for `checkbox` and `radio` input types


packages/conform-react/helpers.ts
```diff
@@ -258,13 +257,16 @@ export function getInputProps<Schema, Options extends InputOptions>(
...
+				typeof options.value === 'string' ? options.value : 'on';
 			props.defaultChecked =
 				typeof metadata.initialValue === 'boolean'
 					? metadata.initialValue
-					: metadata.initialValue === props.value;
+					: metadata.initialValue === props.defaultValue;
 		} else if (typeof metadata.initialValue === 'string') {
 			props.defaultValue = metadata.initialValue;
+		} else if (typeof metadata.initialValue === 'number') {
+			props.defaultValue = metadata.initialValue.toString();
```
lifeiscontent, line 269: @edmundhung one thing that I was questioning, is should this even be `initialValue`? it seems `value` already does the casting for us via `serialize`

edmundhung: I think `serialize` should cast everything to `string` already. So this is probably not needed.

playground/app/routes/collection.tsx
```diff
@@ -45,9 +45,9 @@ export default function Example() {
 						type: 'radio',
 						options: ['x', 'y', 'z'],
 					}).map((props) => (
-						<label key={props.value} className="inline-block">
+						<label key={props.defaultValue} className="inline-block">
 							<input {...props} />
-							<span className="p-2">{props.value?.toUpperCase()}</span>
+							<span className="p-2">{props.defaultValue.toUpperCase()}</span>
```
edmundhung, line 50: This is definitely a breaking change now unless we preserve `props.value` 🤔

packages/conform-react/helpers.ts
lifeiscontent: if we use `initialValue` its not a string, its the the fully represented type I have patched this functionality locally because previously it was returning nothing for values that were numbers.

lifeiscontent: and also from the #613 PR, we should probably also serialize `bigint` here too

playground/app/routes/collection.tsx
lifeiscontent: to be honest, I have some ideas on how to improve the collection helper, e.g. sometimes you want the `item` rather than just the `value` so you can provide better labels or more info like a badge or something, but not sure if that's worth doing now.

lifeiscontent: @edmundhung would you want to chat about making these helpers more flexible? I could also create a gist of some ideas if you don't have time to discuss at the moment.

edmundhung: I would avoid a breaking change unless it has huge impact at this stage. At least not when v1 is just released 3 months ago 😅 But I would still be happy to learn more about your ideas even there is breaking changes. I am also considering getting rid of all these exports in v2 and provide a CLI that generate these helpers similar to shadcn-ui. Then we don't need to care about breaking changes on these helpers anymore.

packages/conform-react/helpers.ts
edmundhung: The `initialValue` you got from the metadata is already serialized. So once #613 is merged, this should be good.

playground/app/routes/collection.tsx
lifeiscontent: @edmundhung yeah, maybe we can have a call on discord and do some brain storming if you're available

edmundhung:
As all helpers are removed in v2, I am gonna close this PR for now. Thanks.

PR #607, opened 2024-04-28
fix: subfield errors should be consider validated based on its parent

edmundhung:
Fix #596


edmundhung:
@lifeiscontent It seems like our [CI fails to start the playground app](https://github.com/edmundhung/conform/actions/runs/8867091799/job/24345266223?pr=607) randomly since we changed the build setup. Let me know if you have any ideas why.

PR #613, opened 2024-04-28
fix: serialize for bigint

lifeiscontent:


packages/conform-dom/submission.ts
```diff
@@ -524,14 +524,11 @@ export function setListState(
 	});
 }
 
-export function serialize<Schema>(
-	defaultValue: DefaultValue<Schema>,
-): FormValue<Schema> {
+export function serialize<Schema>(defaultValue: Schema): FormValue<Schema> {
```
lifeiscontent, line 527: @edmundhung the reason for this change is because the call sites for `serialize` already are wrapped with `DefaultValue<Schema>` so in use it would be wrapped again with `DefaultValue<DefaultValue<Schema>`

edmundhung (approved):
LGTM

PR #615, opened 2024-04-29
chore: fix playground build

lifeiscontent:
@edmundhung I'm hoping this will fix the issues we're seeing in playwright


guide/package.json
```diff
@@ -33,9 +33,9 @@
 		"eslint": "^8.35.0",
 		"tailwindcss": "^3.4.0",
 		"typescript": "^5.3.3",
-		"wrangler": "^3.22.1"
+		"wrangler": "^3.28.2"
```
lifeiscontent, line 36: running `pnpm dedupe` said these versions were the correct requested versions from `@remix-run/dev`

playground/tailwind.config.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export default {
+	content: ['./app/**/*.{js,jsx,ts,tsx}'],
+	theme: {
+		extend: {},
+	},
+	plugins: [
+		form,
+		plugin(function ({ addVariant }) {
+			addVariant('touched', '&[data-conform-touched="true"]');
```
lifeiscontent, line 13: @edmundhung is this an artifact from previous versions of conform? maybe we can remove?

edmundhung: We can remove it. It's a legacy thing pre v1.

edmundhung (dismissed):
Thanks! I think the CI issue should be already resolved with #614, but it's great to update the playground app anyway.

edmundhung (approved):
Thanks! :)

PR #616, opened 2024-04-29
chore: make tests OS agnostic

lifeiscontent:
### What I did

1. Update Playwright to run tests across MacOS and Linux
2. Update Playwright to run across 4 different shards to speed up test runtime
3. Update Playwright to use `./tests/integration` as when using VSCode is provides a nice icon to the `integration` folder if using an icon pack
4. Fix Playwright report merging
5. Update tests to handle keyboard interactions across various OS types
6. Use corepack in CI to install `pnpm`
7. Use `process.env.NODE_ENV === 'production'` to determine if React is running in Strict mode rather than `process.env.CI`
8. add favicon to playground to stop showing 404 errors in playwright
9. fix typo of `isStrictMode`

closes #379


tests/integrations/subscription.spec.ts
```diff
@@ -48,244 +48,489 @@ async function runTest(page: Page) {
...
+			'message.id: 1',
+			'message.formId: 1',
+			'message.errorId: 1',
+			'message.descriptionId: 1',
+			'message.initialValue: 1',
+			'message.value: 1',
+			'message.key: 1',
+			'message.dirty: 1',
+			'message.valid: 1',
+			'message.errors: 1',
```
lifeiscontent, line 82: @edmundhung I'm not sure if this is concerning to you or not, but the values are different between OS's if you have an idea on how to make this more deterministic I'm open to hearing your feedback.

tests/integrations/async-validation.spec.ts
```diff
@@ -74,7 +74,11 @@ async function runTest(page: Page, javaScriptEnabled: boolean) {
 		'Title is required',
 	]);
 
-	await fieldset.email.press('Control+a');
+	if (process.platform === 'darwin') {
+		await fieldset.email.press('Meta+a');
+	} else {
+		await fieldset.email.press('Control+a');
+	}
```
edmundhung, line 81: Maybe we can select the whole field with a [double click](https://playwright.dev/docs/api/class-locator#locator-dblclick)? Then we don't need OS specific approach.

edmundhung (commented):
Thanks for working on this!

Should we run the same test suites on window as well? Also, let's not change any package code in this PR.

tests/integrations/subscription.spec.ts
edmundhung: This seems to be fixed? I guess you might had issue with the strict mode setup.

lifeiscontent: @edmundhung actually running locally on my machine, the results differ from CI, not sure why

edmundhung: Those number are the render count and will be different dependings whether you are running react in dev with strict mode (e.g. double renders) or in production. To get a stable number, I did some [simple math](https://github.com/edmundhung/conform/blob/be32a14669dc62fd8a470b062c80a18b027068e7/playground/app/routes/subscription.tsx#L136-L148) here to normalize the value based on whether the strict mode is running. This is decided by the CI env [here](https://github.com/edmundhung/conform/blob/be32a14669dc62fd8a470b062c80a18b027068e7/playground/app/routes/subscription.tsx#L40). 

I usually run the playground in dev mode locally so the env doesn't matter. But if you run the test without a running playground, it will run the playground in production mode without the CI env. That's why the number wrong.

tests/integrations/async-validation.spec.ts
lifeiscontent: @edmundhung there are a few other cases where you copy/paste values as well, so I opt-ed for helper functions instead

lifeiscontent:
@edmundhung I'm not sure if its worth it, the E2E tests with MacOS + Windows increase by more than double in terms of execution time, if you want to proceed with it anyway, I'll update to run MacOS again I tried to get Windows to run, but it looks like the playground Remix app doesn't run on windows currently 😅

package.json
```diff
@@ -35,6 +35,7 @@
 	"engines": {
 		"node": "20.x"
 	},
+	"packageManager": "pnpm@9.1.0+sha256.22e36fba7f4880ecf749a5ca128b8435da085ecd49575e7fb9e64d6bf4fad394",
```
lifeiscontent, line 38: @edmundhung this field allows us to use `corepack` so now in CI or development, you can just run `corepack enable` and it'll automatically use the correct package manager.

to update you can run `corepack up` and that will also rewrite the hash for the version number like it is here.

playwright.config.ts
```diff
@@ -10,17 +9,8 @@ import { devices } from '@playwright/test';
 /**
  * See https://playwright.dev/docs/test-configuration.
  */
-const config: PlaywrightTestConfig = {
-	testDir: './tests/integrations',
```
edmundhung, line 14: I would prefer keeping test files inside `tests`

package.json
edmundhung: Do you recommend corepack? I think I have tried it once but there were some issue with pnpm so I didnt use it at the end. If you suggest this, I can give it another try.

edmundhung (changes requested):
Just one minor issue with the folder structure changes. Otherwise looks good to me. Thank you!

edmundhung:
> @edmundhung I'm not sure if its worth it, the E2E tests with MacOS + Windows increase by more than double in terms of execution time, if you want to proceed with it anyway, I'll update to run MacOS again I tried to get Windows to run, but it looks like the playground Remix app doesn't run on windows currently 😅

We can run them in parallel with GitHub action, right? I am fine to merge it as is but it would definitely be a big plus to be able to test on all 3 platforms.

package.json
lifeiscontent: @edmundhung I use it on my project at work, I'd love to hear about the issues you had, with the setup I've done here, I haven't personally witnessed any issues.

lifeiscontent:
@edmundhung this is ready for review now, just need to update the required status checks in GitHub's Settings to the new names

lifeiscontent:
@edmundhung using Sharding in Playwright the total test time has slightly increased by a few seconds, so you won't see too much of a longer wait 🎉

edmundhung (approved):
Thank you!

PR #633, opened 2024-05-08
chore: add typecheck and fix type

lifeiscontent:


packages/conform-react/context.tsx
```diff
@@ -83,7 +71,7 @@ type SubfieldMetadata<
 	Schema,
 	FormSchema extends Record<string, any>,
 	FormError,
-> = [Schema] extends [Primitive]
+> = [Schema] extends [Primitive | Date | File]
```
lifeiscontent, line 74: @edmundhung this seems to be the only place Primitive is being used, whats the intent for overloading the term "Primitive" here with `Date` and `File`?

maybe another term like `FormValue` or something along those lines might be better here

packages/conform-dom/form.ts
```diff
@@ -623,7 +623,7 @@ function shouldNotify<Schema>(
 	return false;
 }
 
-export function createFormContext<
```
edmundhung, line 626: What does this helps comparing to setting an alias on the entry file?

packages/conform-dom/package.json
```diff
@@ -18,12 +18,13 @@
 	},
 	"scripts": {
 		"build:js": "rollup -c",
-		"build:ts": "tsc",
+		"build:ts": "tsc --emitDeclarationOnly",
```
edmundhung, line 21: Why do we need to remove `emitDeclarationOnly` from the tsconfig?

packages/conform-react/context.tsx
```diff
@@ -272,24 +260,6 @@ export function getMetadata<
...
-									stateSnapshot,
-									name,
-									key,
-								);
-							}
-
-							return Reflect.get(target, key, receiver);
-						},
-					});
-			},
```
edmundhung, line 292: Why?

packages/conform-react/integrations.ts
```diff
@@ -294,10 +294,12 @@ export function useInputValue<
 	return [value, setValue] as const;
 }
 
-export function useControl<
+export function unstable_useControl<
 	Value extends string | string[] | Array<string | undefined>,
 >(meta: { key?: Key | null | undefined; initialValue?: Value | undefined }) {
+	// eslint-disable-next-line react-hooks/rules-of-hooks
 	const [value, setValue] = useInputValue(meta);
+	// eslint-disable-next-line react-hooks/rules-of-hooks
```
edmundhung, line 302: Err.. why?

edmundhung (commented):
Can you share some context the usecase with the typecheck script? A lot of the changes here seems styling changes only.

packages/conform-dom/form.ts
lifeiscontent: you as the author know its unstable, but as the contributor I might not if I don't look at the export boundary (index) that I would read this as stable code

packages/conform-dom/package.json
lifeiscontent: when running `tsc` with `--noEmit` and `--emitDeclarationOnly` is set in the tsconfig, `tsc` throws an error saying only one can be set at a time.

packages/conform-react/context.tsx
lifeiscontent: @edmundhung based off the type it doesn't seem to be used, is it? tests still seem to pass.

packages/conform-react/integrations.ts
lifeiscontent: since the name of the function doesn't start with `use` it complains, I'm not sure what remix team does with thier unstable APIs to get around this but I could look into it. I'm guessing they might have lint rules turned off for things marked as `unstable_`

edmundhung:
This has gone really stale and there's not much we can action from these changes now. I'm going to close this PR. Thank you.

PR #671, opened 2024-06-07
fix: TypeError on creating properties on immutable objects

### LINKED ISSUE #670: Found TypeError: Cannot create property on string

If we first pass an immutable object in the url parameters and then try to add a property to it, we get the `TypeError: Cannot create property on string.` I expect that the library will not return an error and will give the first values it finds when unpacking.

### Conform version

v1.1.4

```
import { parseWithZod } from '@conform-to/zod'
import { z } from 'zod'

const someSchema = z.object({
 someNumber: z.number().min(1).optional(),
 someString: z.string().min(1).optional(),
})

const url = new URL('http://localhost/smth?c=name&c.A=somea')
const parsed = parseWithZod(url.searchParams, {
    schema: someSchema,
})

console.log(parsed)```

_No response_

### Screenshots or Videos

_No response_

### Additional context

Openning PR for fixing issue

### PR DESCRIPTION
**Pull request by novikovfred**
fix #670

### DIFF
--- a/packages/conform-dom/formdata.ts
+++ b/packages/conform-dom/formdata.ts
@@ -129,9 +129,9 @@ export function setValue(
 				: valueFn(pointer[key]);
 		if (
 			Object.isExtensible(pointer) &&
-			!!nextKey &&
-			!(Object.isSealed(pointer) && !(nextKey in pointer)) &&
-			!(Object.isFrozen(pointer) && !(nextKey in pointer))
+			!!key &&
+			!(Object.isSealed(pointer) && !(key in pointer)) &&
+			!(Object.isFrozen(pointer) && !(key in pointer))
 		) {
 			pointer[key] = newValue;
 		}
--- a/packages/conform-dom/formdata.ts
+++ b/packages/conform-dom/formdata.ts
@@ -127,16 +127,36 @@ export function setValue(
 						? []
 						: {}
 				: valueFn(pointer[key]);
-		if (
-			Object.isExtensible(pointer) &&
-			!!key &&
-			!(Object.isSealed(pointer) && !(key in pointer)) &&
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: packages/conform-dom/formdata.ts [hunk 1]
```diff
@@ -128,11 +128,37 @@ export function setValue(
...
+	}
+	if (Object.isSealed(obj) && !(propertyStr in obj)) {
+		return false;
+	}
+	if (Object.isFrozen(obj) && !(propertyStr in obj)) {
+		return false;
+	}
+	return true;
+}
+
```
edmundhung, line 161: I wanna understand better the impact of this change. Can you explain a bit more what each condition is trying to handle?

novikovfred: As i mentioned in #670 currently we can get TypeError if we will try to assign property on immutable object. 
Since we validate and get data from those parameters that can be passed by users, including frauders. We wouldn't want to get errors in runtime because of a user who enters intentionally incorrect data. Of course, I can handle the error, but I would like not to get errors in the application's runtime.

novikovfred: The same issue we can reproduce with formdata as well.

confix:
Actually the check makes perfect sense to me. We were bitten by that today as well. 

Someone updated a fieldset (object) via `form.update` with a string value instead of an object.
Afterwards we want to set a property of the mentioned fieldset via a regular input field and the error occurs.

```javascript
// state: { fieldset: { property: "aValue" }}
form.update({ name: "fieldset", value "a string" })
// state: { fieldset: "a string" }

// a change to the input below crashes with the same error
<input name="fieldset.property" />
```

We should report a warning as long as we're in development, though. We could end up in the very same situation by simply having a bug in our application.

PR #674, opened 2024-06-11
fix: reset registered input value on key changes

### LINKED ISSUE #673: [1.1.0] `resetForm: true` doesn't reset `unstable_useControl` values

As of Conform v1.1.0 (and through v1.1.4), resetting the form doesn't seem to reset `unstable_useControl` values (unless, possibly, it's the only field in the form - see repro).

### Conform version

v1.1.4

https://stackblitz.com/edit/remix-run-remix-pvh5h7?file=app%2Froutes%2F_index.tsx

Chrome, Microsoft Edge

### Screenshots or Videos

_No response_

### Additional context

_No response_

edmundhung:
Fix #673

### DIFF
--- a/.changeset/clean-spies-clap.md
+++ b/.changeset/clean-spies-clap.md
@@ -1,0 +1,5 @@
+---
+"@conform-to/react": patch
+---
+
+fix: reset registered input value on key changes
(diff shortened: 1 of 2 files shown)


halljus:
How are we feeling about this patch? 🙂

PR #686, opened 2024-06-18
chore: add json example

lifeiscontent:
@edmundhung can you explain to me how I might be able to get a JSON field working with conform? I've setup this example, but the `.defaultValue` of the input renders as a blank input value

### DIFF
--- a/playground/app/routes/parse-with-zod.tsx
+++ b/playground/app/routes/parse-with-zod.tsx
@@ -1,0 +1,74 @@
+import { getFormProps, getInputProps, useForm } from '@conform-to/react';
+import { parseWithZod } from '@conform-to/zod';
+import {
+	type LoaderFunctionArgs,
+	type ActionFunctionArgs,
+	json,
+} from '@remix-run/node';
+import { Form, useActionData, useLoaderData } from '@remix-run/react';
+import { Playground, Field } from '~/components';
+import { z } from 'zod';
+
+const literalSchema = z.union([z.string(), z.number(), z.boolean(), z.null()]);
+type Literal = z.infer<typeof literalSchema>;
+type Json = Literal | { [key: string]: Json } | Json[];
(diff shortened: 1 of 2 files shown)


edmundhung:
Just pushed a commit to your PR. 

There is no JSON field in Conform. Conform only understand string or File in an input. You will need to manually serialize it between string and JSON and make sure you serialize the value again when you update the field value (e.g. `control.change(JSON.serialize(value))`)

lifeiscontent:
@edmundhung is there anyway to use the jsonSchema type I had? for my use case I want to allow someone to add any JSON data because we're mapping it back to a users DB so we don't know the shape of the data.

edmundhung:
I have tested it once with your original type and it seems to recursively looping over and over again and crash. I am not sure why. It might be either a bug on the `enableTypeCoercion` helper in the conform zod package or something wrong with zod itself? Can you verify if the JSON schema works without passing to Conform?

lifeiscontent:
@edmundhung yep the jsonSchema type is in the GitHub docs of zod and works as expected.

edmundhung:
@lifeiscontent I have pushed up a fix to your branch. Let me know if this is what you were looking for.

lifeiscontent:
@edmundhung is there a way I can try a prerelease version of this quickly? I can try it in the project I'm working in and get back to you.

edmundhung:
> @edmundhung is there a way I can try a prerelease version of this quickly? I can try it in the project I'm working in and get back to you.

Here you go :)

lifeiscontent:
@edmundhung it works perfectly! :)

lifeiscontent:
@edmundhung any chance the work you did here could get merged?

edmundhung:
@lifeiscontent Just curious if #871 solves what you need, or is there anything else you want there. Let me know.

lifeiscontent:
@edmundhung yes, I think it should be fine

PR #704, opened 2024-07-11
fix: applies shouldDirtyConsider to both sides of dirty check

### LINKED ISSUE #703: Potential Bug with shouldDirtyConsider and a Disabled Field

Hello! It seems that `shouldDirtyConsider` is not being correctly applied when a field has been disabled.

I've pasted a Stackblitz below to show the problem. `form.dirty` is immediately `true` even though nothing else has changed and the `shouldDirtyConsider` is set to return `true` for any field other than `email`.

Thanks for reading!

### Conform version

v1.1.5

[Stackblitz](https://stackblitz.com/edit/remix-run-remix-gel12n?file=app%2Froutes%2F_index.tsx)

_No response_

### Screenshots or Videos

_No response_

### Additional context

_No response_

ericsakmar:
fixes #703

### DIFF
--- a/packages/conform-dom/form.ts
+++ b/packages/conform-dom/form.ts
@@ -572,25 +572,29 @@ function createDirtyProxy(
 	value: Record<string, unknown>,
 	shouldDirtyConsider: (name: string) => boolean,
 ): Record<string, boolean> {
-	return createStateProxy(
-		(name) =>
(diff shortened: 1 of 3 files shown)


edmundhung:
Closing now as we believe `readonly` attribute is a solution.

PR #706, opened 2024-07-14
Replace `unknown` in submission payload

### LINKED ISSUE #628: Remix single fetch: returning submission.reply() with action results in type never.

https://github.com/remix-run/remix/issues/9374

### Conform version

v1.1.3

https://stackblitz.com/~/github.com/coji/conform-on-remix-single-fetch

Chrome

### Screenshots or Videos

CleanShot 2024-05-04 at 23 12 53@2x

### Additional context

I have also raised it as a remix issue. I think it may be a type definition problem on the remix side.
https://github.com/remix-run/remix/issues/9374

timvandam:
Fix #628

This pull request replaces the `unknown` type in submission payloads. `unknown` is not compatible with Remix Single Fetch and new `defineLoader` or `defineAction` as it cannot be guaranteed that values of type `unknown` can be sent over the wire.

Removing `unknown` types addresses this issue. I've replaced it with what I think is correct but I don't know the ins and outs of this library so plz check.

Thanks!

### DIFF
--- a/.changeset/gentle-humans-eat.md
+++ b/.changeset/gentle-humans-eat.md
@@ -1,0 +1,5 @@
+---
+'@conform-to/dom': patch
+---
+
+Replace submission payload unknown types
--- a/packages/conform-dom/submission.ts
+++ b/packages/conform-dom/submission.ts
@@ -14,9 +14,14 @@ export type SubmissionState = {
 	validated: Record<string, boolean>;
 };
 
+export type SubmissionPayload<Entry extends FormDataEntryValue> =
+	| Entry
+	| SubmissionPayload<Entry>[]
+	| { [key: string]: SubmissionPayload<Entry> };
+
 export type SubmissionContext<Value = null, FormError = string[]> = {
 	intent: Intent | null;
-	payload: Record<string, unknown>;
+	payload: Record<string, SubmissionPayload<FormDataEntryValue>>;
 	fields: Set<string>;
 	value?: Value;
 	error?: Record<string, FormError | null> | null;
@@ -26,13 +31,13 @@ export type SubmissionContext<Value = null, FormError = string[]> = {
 export type Submission<Schema, FormError = string[], FormValue = Schema> =
 	| {
 			status: 'success';
-			payload: Record<string, unknown>;
+			payload: Record<string, SubmissionPayload<FormDataEntryValue>>;
 			value: FormValue;
 			reply(options?: ReplyOptions<FormError>): SubmissionResult<FormError>;
 	  }
 	| {
 			status: 'error' | undefined;
-			payload: Record<string, unknown>;
+			payload: Record<string, SubmissionPayload<FormDataEntryValue>>;
 			error: Record<string, FormError | null> | null;
 			reply(options?: ReplyOptions<FormError>): SubmissionResult<FormError>;
 	  };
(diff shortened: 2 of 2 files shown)


timvandam:
This does not seem to work for some playground files because `File` cannot be JSONified, breaking `useActionData`. While `File` is never sent from server to client, this is not reflected in the type of `SubmissionResult.reply()`. What do you think of adding a new reply option that narrows the type to exclude files in the payload type? I think this would make types work with Remix Single Fetch/defineLoader/defineAction using syntax like `result.reply({ excludeFiles: true })`

Let me know if that works and I will change this PR @edmundhung

edmundhung:
Thanks for the PR @timvandam. I was not aware that I can represent the submission payload this way, but now I see you did that it make total sense :+1:

> What do you think of adding a new reply option that narrows the type to exclude files in the payload type? I think this would make types work with Remix Single Fetch/defineLoader/defineAction using syntax like result.reply({ excludeFiles: true })

We are stripping all files already if you are [running on the server](https://github.com/edmundhung/conform/blob/main/packages/conform-dom/submission.ts#L284-L285). So to make it non-breaking, we will need to default the new option to `true`. But I will just exclude File type on the `SubmissionResult` for now since only Conform's internal care about the File and it is something I plan to change soon.

I can't push any changes to your PR likely because your PR was based on your main branch. This is what I am thinking:

```tsx
export type SubmissionPayload<Entry extends FormDataEntryValue> =
	| Entry
	| SubmissionPayload<Entry>[]
	| { [key: string]: SubmissionPayload<Entry> };

export type Submission<Schema, FormError = string[], FormValue = Schema> =
	| {
			status: 'success';
			payload: Record<string, SubmissionPayload<FormDataEntryValue>>;
			value: FormValue;
			reply(options?: ReplyOptions<FormError>): SubmissionResult<FormError>;
	  }
	| {
			status: 'error' | undefined;
			payload: Record<string, SubmissionPayload<FormDataEntryValue>>;
			error: Record<string, FormError | null> | null;
			reply(options?: ReplyOptions<FormError>): SubmissionResult<FormError>;
	  };

export type SubmissionResult<FormError = string[]> = {
	status?: 'error' | 'success';
	intent?: Intent;
	initialValue?: Record<string, SubmissionPayload<string>> | null;
	fields?: string[];
	error?: Record<string, FormError | null>;
	state?: SubmissionState;
};
```

Then, on `replySubmission`, we will do this:
```tsx
const initialValue = normalize(
    context.payload,
    // We can't serialize the file and send it back from the server, but we can preserve it in the client
    typeof document !== 'undefined',
    // We need the file on the client because it's treated as the form value
    // But we will exclude the File type for now as it's only used by the internal
    // form state and we will remove the need to preserve the file on the client soon
) as Record<string, SubmissionPayload<string>> ?? {}

return {
    initialValue,
    // ...
};
```

timvandam:
Awesome, not sure why you cannot push to my branch, but I've added what you've commented. Very nice solution with the generic, I was thinking of much more complex ways of going about it

edmundhung (approved):
Love this change. Thanks!

JasonColeyNZ:
Hi guys, when will this PR be released, this is very much needed right now for Remix and Single Fetch?

edmundhung:
I am hoping to cut a release sometimes this month. If you can give the pre-release version a try and let me know whether you run into any issues, it would help me releasing this change sooner. Thanks!

JasonColeyNZ:
> I am hoping to cut a release sometimes this month. If you can give the pre-release version a try and let me know whether you run into any issues, it would help me releasing this change sooner. Thanks!

Sorry to sound dim, how do I do this, do I just change my package .json to ...

```
    "@conform-to/react": "pre-release",
    "@conform-to/zod": "pre-release",
```

edmundhung:
There is a comment above from pkg.pr.new with each package name listed. You will find a command to install the pre-release version of that package if you expand the item.

If you are not using pnpm, you can swap it with npm install, or just copy the package URL and replace the version in the package.json with it. 

Really appreciate for giving this a try! 👍

PR #708, opened 2024-07-19
docs: Fix typo in parseWithYup.md

uttk:
The module name of the import destination was incorrect, so I fixed it.


edmundhung (approved):
How can I not aware of this for so long 😅 Thanks!!

PR #721, opened 2024-07-26
Optimize Shadn Switch Component example by Using control.value

reborn2135:
# Summary
This PR optimizes the Switch component in the shadcn example by replacing meta.value with control.value. This change aims to improve performance by reducing unnecessary re-renders.

# Changes Made
Updated the Switch component to use control.value instead of meta.value
Modified the relevant code snippet as follows:

```tsx
Copy// Before
checked={meta.value === "on"}

// After
checked={control.value === "on"}
```


edmundhung (approved):
Thank you!

PR #726, opened 2024-08-02
feat(conform-zod): bigint coercion support

lifeiscontent:
this change follows the same pattern as how booleans are coerced.

it also fixes issues with BigInt where if the value is not of a convertable value it doesn't break the app because BigInt/Symbol throw unlike the rest of the type constructors.

### DIFF
--- a/tests/conform-zod.spec.ts
+++ b/tests/conform-zod.spec.ts
@@ -405,6 +405,74 @@ describe('conform-zod', () => {
 			});
 		});
 
+		test('z.bigint', () => {
+			const schema = z.object({
+				test: z
+					.bigint({ required_error: 'required', invalid_type_error: 'invalid' })
+					.min(1n, 'min')
+					.max(10n, 'max')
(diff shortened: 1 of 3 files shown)


tests/conform-zod.spec.ts
```diff
@@ -400,7 +400,7 @@ describe('conform-zod', () => {
 			).toEqual({
 				status: 'error',
 				payload: { test: ' ' },
-				error: { test: ['invalid'] },
```
edmundhung, line 403: This is exactly what we are trying to avoid with `.trim()`

edmundhung (changes requested):
Thanks for the PR! The changes for BigInt makes sense to me but maybe we should keep the number coercion untouched.

lifeiscontent:
@edmundhung fixed

edmundhung (dismissed):
Looks good to me. Thank you!

edmundhung (approved):

PR #728, opened 2024-08-03
feat: update field value automatically

### LINKED ISSUE #600: Warning: A props object containing a "key" prop is being spread into JSX:

I am getting a console error in Next when using conforms `getInputProps`.

### Conform version

1.1.2

```
<input {...getInputProps(fields.id, { type: 'hidden' })} />
```

I get the error message 

```
Warning: A props object containing a "key" prop is being spread into JSX:
  let props = {key: someKey, id: ..., name: ..., form: ..., type: ...};
  <input {...props} />
React keys must be passed directly to JSX without using spread:
  let props = {id: ..., name: ..., form: ..., type: ...};
  <input key={someKey} {...props} />
```

_No response_

### Screenshots or Videos

_No response_

### Additional context

_No response_

edmundhung:
Fix #600.

This is an attempt to remove the need of passing the `key` to each input manually by updating the fields value automatically. It means:

1. No more warnings from React when spreading the result of `getInputProps()`, `getSelectProps()` or `getTextareaProps()` as it will no longer include a `key`.
2. The `name` will become the only metadata required to be set on your inputs. The `initalValue` is required only if you care about progressive enhancement.

The main challenge of this change is that Conform will trigger a re-render only when the subscribed metadata is changed and so we don't have a single place that will always re-render and runs the side effect we need. This PR resolves it by running the side effect everywhere we made a subscription and update the relevant fields only if the `initialValue` is subscribed.


edmundhung:
Superseded by #729

PR #734, opened 2024-08-09
fix: getInputProps

lifeiscontent:


edmundhung:
> the file input will have an array of undefined values and [undefined,undefined]

I guess this happens only with a multiple file input on full document reload in which we have to strip the files from the server as it is not serializable. 

But I am not sure how it ends up with the value you see there because the initialValue would still be array so it shouldn't see anything to the default value. Do I missed something?

lifeiscontent:
> Do I missed something?

Yeah, I forget which test I came across that was showing it, but if you throw an error in the `getInputProps` by checking if its an array you'll find it if you run playwright tests

Otherwise, I can follow up when I find it again.

PR #741, opened 2024-08-15
fix(conform-zod): empty string default value support

### LINKED ISSUE #676: schema not allowing empty string through

https://stackblitz.com/edit/remix-run-remix-fd6e6f?file=app%2Froutes%2F_index.tsx,app%2Fentry.client.tsx

I'd expect being able to press "Submit" to work because the default value of inputs is `""`

### Conform version

v.1.1.4

1. Goto the stackblitz page
2. press enter
3. see validation error instead of submission

_No response_

### Screenshots or Videos

_No response_

### Additional context

_No response_

edmundhung:
Fix #676

### DIFF
--- a/docs/api/zod/parseWithZod.md
+++ b/docs/api/zod/parseWithZod.md
@@ -86,9 +86,7 @@ const schema = z.object({
 
 ### Default values
 
-Conform already preprocesses empty values to `undefined`. Add `.transform()` to your schema to define a default value that will be returned instead.
-
-Zod will return the default value if the input is `undefined` after preprocessing. This also has the effect of changing the schema return type.
+Conform will always strip empty values to `undefined`. If you need a default value, please use `.transform()` to define a fallback value that will be returned instead.
 
 ```tsx
 const schema = z.object({
--- a/docs/api/zod/parseWithZod.md
+++ b/docs/api/zod/parseWithZod.md
@@ -86,14 +86,20 @@ const schema = z.object({
 
 ### Default values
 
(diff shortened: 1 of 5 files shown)


docs/api/zod/parseWithZod.md
```diff
@@ -86,14 +86,20 @@ const schema = z.object({
...
 ```tsx
 const schema = z.object({
-  foo: z.string(), // string | undefined
-  bar: z.string().default('bar'), // string
-  baz: z.string().nullable().default(null), // string | null
+  foo: z.string().optional(), // string | undefined
+  bar: z
+    .string()
+    .optional()
+    .transform((value) => value ?? ''), // string
```
lifeiscontent, line 99: I would expect it work work like `z.string().optional().default("")`

docs/api/zod/parseWithZod.md
```diff
@@ -86,14 +86,20 @@ const schema = z.object({
...
+  baz: z
+    .string()
+    .optional()
+    .transform((value) => value ?? null), // string | null
```
lifeiscontent, line 103:
```
z.string().nullable().default(null)
```

docs/api/zod/parseWithZod.md
edmundhung: It will work after this change. But it is no longer what I recommend to keep the field value an empty string because of how default value will be validated when there is extra checks. 

See the test I added which showcase the issue.

docs/api/zod/parseWithZod.md
edmundhung: Sure this works as well but I wanna keep it consistent.

PR #750, opened 2024-08-22
Update intent-button.md, use `getInputProps`

willhoney7:
I was about to open up a bug report because I couldn't get the `insert` intent to work correctly when setting a `defaultValue`, but then I realized it's because the input needs to use `getInputProps`. I had copied from the example here, so here's a simple PR to update the example to use `getInputProps`.

update: The complex structures page should probably receive some edits as well. https://conform.guide/complex-structures

### DIFF
--- a/docs/intent-button.md
+++ b/docs/intent-button.md
@@ -135,7 +135,7 @@ Be aware that both intents requires setting up the inputs with the **key** from 
 To manipulate a field list, you can use the **insert**, **remove** and **reorder** intents.
 
 ```tsx
-import { useForm } from '@conform-to/react';
+import { useForm, getInputProps } from '@conform-to/react';
 
 export default function Tasks() {
   const [form, fields] = useForm();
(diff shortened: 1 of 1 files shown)


edmundhung:
Hi @willhoney7, sorry for the confusion. If you need to be able to update a field, the minimal setup as of v1.1.5 would be `<input key={task.key} name={task.name} defaultValue={task.initialValue} />` which is what `getInputProps` does for you.

Having said that, this is no longer true once #729 is landed, in which `<input name={task.name} />` is enough as long as you don't care about full document reload (progressive enhancement).

There is a pre-release version in the PR above. Feel free to give it a try and let me know if you run into any problems.

I am likely gonna leave this unmerged so I don't need to revert it later. But I will keep the PR open until the change is landed. Thanks again for putting this up!

edmundhung:
[v1.2.0](https://github.com/edmundhung/conform/releases/tag/v1.2.0) is finally out. Sorry again for the confusing example there. I hope you like the new changes :)

Issue #756, opened 2024-08-30
resetForm: true resets status


lifeiscontent:
### Describe the bug and the expected behavior

I'm not sure if I'm misunderstanding, but I have a scenario where if the form submission was unsuccessful I want to keep a form drawer open, so my plan was to check the submission status, and if it wasn't equal to `success` keep the form drawer open so the user can take care of the errors

it appears that when we return `json(submission.reply({resetForm: true}))` it actually clears out the status, how can I be sure that the submission was successful?

### Conform version

v1.1.5

### Steps to Reproduce the Bug or Issue

1. Create a Form Drawer
2. use a fetcher
3. in the action do what I described above
4. see that if the action was successful the status is not defined in the `fetcher.data`

Garajluc:
Any updates on this?

lifeiscontent:
@Garajluc yes, but what's your use case? I can give some guidance depending on what you need because it'll vary depending on what your expected outcome is.

Garajluc:
Thank you for your response, I actually read through similar issues within the conform repository and I think I understand it better now, so no need for additional guidance at this moment. Thank you again :)

PR #757, opened 2024-08-30
Update submission.ts

lifeiscontent:
fixes: #756

### DIFF
--- a/packages/conform-dom/submission.ts
+++ b/packages/conform-dom/submission.ts
@@ -247,6 +247,21 @@ export function replySubmission<FormError>(
 		(context.intent?.type === 'reset' &&
 			(context.intent.payload.name ?? '') === '')
 	) {
+		const extraError =
+			'formErrors' in options || 'fieldErrors' in options
+				? normalize<Record<string, FormError | null>>({
+						'': options.formErrors ?? null,
+						...options.fieldErrors,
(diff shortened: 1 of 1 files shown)


packages/conform-dom/submission.ts
```diff
@@ -247,7 +247,25 @@ export function replySubmission<FormError>(
...
+						...options.fieldErrors,
+					})
+				: null;
+		const error =
+			context.error || extraError
+				? {
+						...context.error,
+						...extraError,
+					}
+				: undefined;
```
lifeiscontent, line 263: @edmundhung I'm not sure if it makes sense to show the error status, but I kept it in because I'm not sure what the expectations are

lifeiscontent:
@edmundhung I can fix tests if you think this is an appropriate change

lifeiscontent:
nevermind, I figured out how to do it.

this is the code that works with `submission.reply({resetForm: true})`

```tsx
  useEffect(() => {
    if (
      fetcher.state === "idle" &&
      fetcher.data &&
      fetcher.data.status !== "error"
    ) {
      onFinish();
    }
  }, [fetcher.data, fetcher.state, onFinish]);
  ```
  
  where `onFinish` should be called in the scenario you want to close a form drawer.

PR #761, opened 2024-09-12
docs: update intent-button.md

KirillSBarsukov:
Added a working example for insert, remove and reorder intents.

### DIFF
--- a/.changeset/chilled-dingos-sniff.md
+++ b/.changeset/chilled-dingos-sniff.md
@@ -1,5 +1,0 @@
----
-"@conform-to/dom": patch
----
-
-fix: FormValue should never be null
--- a/.changeset/chilled-dingos-sniff.md
+++ b/.changeset/chilled-dingos-sniff.md
@@ -1,0 +1,5 @@
+---
+"@conform-to/dom": patch
+---
+
+fix: FormValue should never be null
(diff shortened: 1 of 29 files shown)


edmundhung:
Hi @KirillSBarsukov, sorry for the confusing example in the docs. If you need to update the value of a field, the minimal setup as of `v1.1.5` would be `<input key={task.key} name={task.name} defaultValue={task.initialValue} />` and so your changes using `getInputProps` does fix the issue you had with the form intents.

Having said that, the situation is a bit different with our newly released [v1.2.0](https://github.com/edmundhung/conform/releases/tag/v1.2.0). As it is no longer required to pass a `key` to the input element. The `defaultValue` is also optional depending on your usecases. The original example is now working as expected without any changes! Please give it a try and let me know if you run into any other problems :)

edmundhung:
The change looks right to me! Thanks a lot and sorry for all the confusions.

edmundhung (dismissed):

edmundhung:
One little request, can you help applying the same change on the ja docs? You can find it here https://github.com/edmundhung/conform/blob/main/docs/ja/intent-button.md

This is pure code change so we don't need to worry about translations. Thanks again :)

edmundhung (approved):
Thank you!

Issue #774, opened 2024-09-16
collection checkboxes not behaving correctly


lifeiscontent:
### Describe the bug and the expected behavior

I'm not sure how to reproduce this but between versions 1.1.5 and 1.2.1 I'm seeing situations where if I have a defaultValue of more than 1 items selected, and have several other options, conform only seems to render 1 item as checked in the form.

### Conform version

v1.2.1

### Steps to Reproduce the Bug or Issue

here's a video of the reproduction:

https://www.loom.com/share/be05ee2475cd40c6b4555469869aef34

### What browsers are you seeing the problem on?

Chrome, Firefox, Microsoft Edge, Safari

tacomanator:
Same issue for me on 1.2.1. Was working as of 1.1.5.

Double checked that the `fields.foo.initialValue` has the expected values, but the corresponding checkboxes are not checked in the page. I noticed that they are checked for a brief moment when the page loads, but become unchecked shortly thereafter.

edmundhung:
I can't watch the loom video without an account. Any chance you can share a simple repo? 😅 I will take a look later today to see if I can figure something out. Sorry for the trouble.

edmundhung:
Looks like loom is complaining only when I have Cloudflare Warp on 🤔 Anyway, I figure out the issue. I will put up a PR with a pre-release for you to test it out.

edmundhung:
I have put up a fix on #775. Can you give the [pre-release](https://github.com/edmundhung/conform/pull/775#issuecomment-2356978130) a try and let me know if you run into any issues? Thanks.

lasseklovstad:
I had the same issue and I tested the pre-relase and it worked for me 😄

edmundhung:
I'm sorry for the trouble. I've concluded that it's best to revert the relevant changes, and I've just published [v1.2.2](https://github.com/edmundhung/conform/releases/tag/v1.2.2).

Issue #771, opened 2024-09-16
Auto-updating the field's value breaks when not using useForm's defaultValue


AndreSilva1993:
### Describe the bug and the expected behavior

Before https://github.com/edmundhung/conform/pull/729 went in I could have something like this:

```
const schema = z.object({ displayName: z.string() });

const [form, fields] = useForm({
  onValidate({ formData }) {
    return parseWithZod(formData, { schema });
  },
});

const defaultValueForInput = "defaultValueForInput";

<Texfield
  name={fields.displayName.name}
  defaultValue={defaultValueForInput}
/>
```

The input was properly initialised with the provided default value and that'd be it. With the https://github.com/edmundhung/conform/pull/729, either we provide a `defaultValue` to the `useForm` hook or the input's value will be overwritten to an empty text input / unchecked radio or checkbox.

I think such a change could either be marked as a **major** upgrade since the documentation / previous versions did not force users to provide the `defaultValue` to the useForm hook when wanting to pass a default value to their inputs - meaning that inputs that are initialised "out of the useForm realm" will be overwritten. Not necessarily arguing that this is better than using the `defaultValue` property in `useForm` since I can see arguments for both side, but it definitely caught me off-guard and broke my edit forms since they would be all empty by bumping to this version.

Or it could make sure that it only overwrites stuff if the `defaultValue` is provided and input is currently empty.

### Conform version

v1.2.1

### Steps to Reproduce the Bug or Issue

```
const schema = z.object({ displayName: z.string() });

const [fields] = useForm({ 
  onValidate: ({ formData }) => {
    return parseWithZod(formData, { schema });
  }
});

const inputDefaultValue = "default-value";

<input name={fields.displayName.name} defaultValue={inputDefaultValue} />
```

In here we can assert that the input will in fact be changed to an empty one.

### What browsers are you seeing the problem on?

Chrome, Firefox, Microsoft Edge, Safari

itsmegood:
Same issue if my isActive boolean default value is set to true the form won't pick up the value until I turn the value off and then on again

edmundhung:
Thanks for the writeup. I have gave it some thoughts last night and my current plan is to make this opt-in with an option to customize the behavior. I will put up a draft PR with more information later today. Sorry for the trouble!

AndreSilva1993:
[...]

Issue #770, opened 2024-09-15
remix-utils/honeypot is not working anymore


davidesigner:
### Describe the bug and the expected behavior

When you use the [Honeypot from remix-utils](https://github.com/sergiodxa/remix-utils?tab=readme-ov-file#form-honeypot) (as explain in the documentation: with `HoneypotProvider` and `HoneypotInputs`), sometimes, the input `from__confirm` value is missing in the formData (while I see a value in the Provider if I add a `console.log`). Sometimes, it's working well... In e2e tests, the problem is always here on my side.

Maybe it's because there's no more `key` added to the input in the last version?

N.B. no problem in v1.1.5.

### Conform version

v1.2.1

### Steps to Reproduce the Bug or Issue

1. Install the epic stack starter project: https://github.com/epicweb-dev/epic-stack
2. Update conform to v1.2.1.
3. Run e2e tests → broken.

### What browsers are you seeing the problem on?

Chrome, Others

### Screenshots or Videos

<img width="627" alt="Screenshot 2024-09-15 at 19 38 09" src="https://github.com/user-attachments/assets/4f563092-b014-4ef8-855d-5a0278299c0b">

- l.204 is clicking on the submit button;
- l.205 check if the next step is displayed (KO). → because of `HoneypotInputs` missing value.

### Additional context

Do you have a solution to this problem? Ideally, I don't want to have to set `defaultValue` for every form that contains `<HoneypotInputs />`.

edmundhung:
Thanks for the report. I wish I realize this earlier. This is definitely not an uncommon case so we will need to address this as soon as possible.

A few solutions I have in mind right now:

1. Add an option to opt out from been managed by conform

This is becoming a breaking change and I hate it.

2. Add a data attribute to opt-in instead

This can be set with the getInputProps so it won't be a breaking change. But I hate it too because it means the minimal setup is not minimal anymore.

3. Automatic opt-in based on the initial value

Conform will only opt in for auto value update if the initial value matches the settings in the useForm hook. This should solve the issue you had but it adds an assumption that the initial value is always set to something on render. If there are any asynchronous works involved and the input has no initial value at first, conform will opt it in since it consider the initial values matches.😬

edmundhung:
I have put up a fix on #775. Can you give the [pre-release](https://github.com/edmundhung/conform/pull/775#issuecomment-2356978130) a try and let me know if you run into any issues? Thanks.

mxp-qk:
[...]

PR #775, opened 2024-09-17
fix(conform-react): stops managing all inputs by default

edmundhung:
Fix #729

Related Issues:
- #770
- #771
- #774

This stops Conform from managing all inputs, except those that are configured with `getInputProps`, `getSelectProps` and `getTextareaProps`. You can customize the behavior by setting `shouldManagerInput` on the `useForm` hook, e.g.

```tsx
function Example() {
    shouldManageInput(element) {
       // To manage all inputs except the csrf-token input
       return element.name !== 'csrf-token';
    }
  });

  // ...
}
```

If we decide to go with this solution, we can also deprecate `shouldDirtyConsider` later as it could be derived based on the result of `shouldManageInput`.

Todos: 
- [ ] Regression tests
- [ ] Docs update

### DIFF
--- a/packages/conform-react/hooks.ts
+++ b/packages/conform-react/hooks.ts
@@ -179,7 +179,8 @@ export function useForm<
 						'checked' in element &&
 						(element.type === 'checkbox' || element.type === 'radio')
 					) {
-						element.checked = get(defaultValue) === element.value;
+						element.checked =
+							getAll(defaultValue)?.includes(element.value) ?? false;
 					} else {
 						element.value = get(defaultValue) ?? '';
 					}
--- a/packages/conform-react/hooks.ts
+++ b/packages/conform-react/hooks.ts
@@ -1,5 +1,5 @@
 import { type FormId, type FieldName } from '@conform-to/dom';
-import { useEffect, useId, useState, useLayoutEffect } from 'react';
+import { useEffect, useId, useState, useLayoutEffect, useRef } from 'react';
 import {
 	type FormMetadata,
 	type FieldMetadata,
@@ -61,6 +61,14 @@ export function useForm<
 			 * Default to `true`.
 			 */
 			defaultNoValidate?: boolean;
+
+			/**
+			 * Define if the input could be updated by conform.
+			 * Default to inputs that are configured using the `getInputProps`, `getSelectProps` or `getTextareaProps` helpers.
+			 */
+			shouldManageInput?: (
+				element: HTMLInputElement | HTMLSelectElement | HTMLTextAreaElement,
+			) => boolean;
 		}
 	>,
 ): [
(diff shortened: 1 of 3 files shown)


packages/conform-react/hooks.ts
```diff
@@ -61,6 +61,14 @@ export function useForm<
...
 			 */
 			defaultNoValidate?: boolean;
+
+			/**
+			 * Define if the input could be updated by conform.
+			 * Default to inputs that are configured using the `getInputProps`, `getSelectProps` or `getTextareaProps` helpers.
+			 */
+			shouldManageInput?: (
+				element: HTMLInputElement | HTMLSelectElement | HTMLTextAreaElement,
+			) => boolean;
```
lifeiscontent, line 71: don't forget buttons! :)

lukasa1993:
would this have any way to go back to `1.1.5` flow ? currently we are using many hidden fields with bunch of names listing them one by one will be extremely tedious

edmundhung:
> would this have any way to go back to `1.1.5` flow ? currently we are using many hidden fields with bunch of names listing them one by one will be extremely tedious

Unless you are configuring the hidden inputs with helpers like `getInputProps`, this PR should make it work like 1.1.5. Conform will only updates inputs configure using those helpers.

You will have access to the dom element, so feel free to do any logic that makes sense to you, e.g. 
```
function Example() {
    shouldManageInput(element) {
       // Ignore all inputs with `address.` as the prefix
       return !element.name.startsWith('address.');
       
       // Ignore all hidden inputs
       return element.type !== 'hidden';
    }
  });

  // ...
}
```

Otherwise, you can always turn off it completely:
```
function Example() {
    shouldManageInput() {
       return false;
    }
  });

  // ...
}
```

packages/conform-react/hooks.ts
edmundhung: Conform does not manage the buttons (i.e. It should not update the button value). Is there something you want to achieve with this option?

lifeiscontent:
@edmundhung this works for me! 👍🏻

edmundhung:
Some people might still consider this a breaking change if they use the helper functions to configure inputs. An alternative approach could be to make this entirely opt-in for v1 and then switch to opt-out in v2. However, I plan to deprecate the helpers, as they don't integrate well with component libraries. I'd prefer to clearly define the properties they should pass to the input and encourage them to create their own helpers to suit their needs.

lifeiscontent:
@edmundhung I'd be open to that, however there should be some dedicated types exposed for things like the return value of `getFieldset()` and `getFieldList()` I personally have had to reach into `@conform-to/dom` to access the `Combine` type for things, it would be nice not to have to do that since I'm only using `@conform-to/react`

packages/conform-react/hooks.ts
lifeiscontent: ah, okay, I was thinking user intent buttons (not conform intent buttons) but if thats outside this scope, please ignore my comment.

edmundhung:
Superseded by #837.

PR #806, opened 2024-10-23
fix: fix document written about complex structure in japanese

suzukisan22:
as you read document, schema key is not same as you get fields of task.


edmundhung (approved):
Thank you!

Issue #828, opened 2024-11-29
Cannot validate nested object fields


oxc:
### Describe the bug and the expected behavior

I'm trying to use `form.validate.getButtonProps` to validate a complete fieldset, i.e. a field that consists only of nested fields.

The validation will produce the errors of the subfields, but it since for the fieldset field no value is submitted, it will never land in the list of validatedFields, and thus not have any errors assigned.

I would have expected that validating a nested object field causes all subfields of this field to be validated, and have their errors reported.

**Suggestion**
When sending a validate intent, always validate all fields for which the field name is a prefix.

Or, make it configurable with a `recursive` flag on the ValidationIntent payload

### Conform version

v1.2.2

### Steps to Reproduce the Bug or Issue

Here is a sandbox that demonstrates the problem:
https://codesandbox.io/p/devbox/smoosh-frost-gf79qd?workspaceId=1702db6b-2042-4a59-957b-ee67bb3c5948

I've added two buttons to the ToDo List example, a "Validate Task" button, which triggers the described behaviour (of not doing anything, basically).
For comparison, there is also a "Validate Task Content" button, which works as expected.

**Steps**
* Click on **Todo list**
* Click on **Add task**
* Click on **Validate Task**
  * I would have expected the task content to become red
* _Click on **Validate Task Content**_
  * _This works, but it validates only one leaf field_

But there is no way to validate an entire field with subfields.

oxc:
I'm cross-linking this StackBlitz example from https://github.com/edmundhung/conform/issues/830#issuecomment-2743719920 because it helps greatly to illustrate the problem.

---

I have created a StackBlitz based on the multi-validate PR (#878) that demonstrates the problem, based on your suggestion to collect the field names from the Form element:

https://stackblitz.com/edit/uzog2zz5?file=src%2Ftodos.tsx

1. Open the Todos List example
2. Add a Task
3. Click the Clear button to enable the Validate button[^1]
4. Click Validate button
5. Note that only the Content field error shows up in the Task's error list
6. Click the Save button at the bottom of the Form
7. Note that now also the Error for the Tags array shows up in the Task's error list

[^1]: perhaps I did something wrong there, but I guess at the time of first rendering, there are no fields in the Form with the tasks' name prefix, which would be another drawback of your suggestion.

PR #830, opened 2024-12-07
Support validate intent to validate a full fieldset (all fields with a given prefix)

oxc:
Proposed fix for #828

Here is a sandbox with the same example from above Issue, where this patch is applied, and a second button is added to the task fieldset that demonstrates the behaviour:
https://codesandbox.io/p/devbox/sad-tree-fdj7hj?workspaceId=ws_3qoruoNJ3FvmBgCfSAACYX

### DIFF
--- a/packages/conform-dom/form.ts
+++ b/packages/conform-dom/form.ts
@@ -332,6 +332,14 @@ function handleIntent<Error>(
 	switch (intent.type) {
 		case 'validate': {
 			if (intent.payload.name) {
+				if (intent.payload.recursive) {
+					const allNames = Object.keys(meta.error).concat(fields ?? []);
+					for (const name of allNames) {
+						if (isPrefix(name, intent.payload.name)) {
+							meta.validated[name] = true;
+						}
+					}
+				}
 				meta.validated[intent.payload.name] = true;
 			} else {
 				setFieldsValidated(meta, fields);
--- a/packages/conform-dom/submission.ts
+++ b/packages/conform-dom/submission.ts
@@ -304,6 +304,7 @@ export type ValidateIntent<Schema = any> = {
 	type: 'validate';
 	payload: {
 		name?: FieldName<Schema>;
+		recursive?: boolean;
 	};
 };
 


edmundhung:
Hi @oxc, thanks for the PR! I can see this solution works for many cases. But the errors object might not have every fields and so there will be some minor inconsistency depending on the validation logic. Ideally, I think we need a different data structure so we can express is as every fields under `x` is validated, except `y`, `z` etc 🤔 

But as a short term solution, how about making the validate intent accept an array of string? We can combine it with a simple utility to look for all the fields under it to achieve the same functionality, like:

```ts
form.validate({
  // This can be a string or an array of string
  name: ['address', 'address.city', ...],
});

form.validate({
  // Use a helper to look up fields from the DOM
  name: getFields(formElement, name),
});

// Example `getFields()` implementation
function getFields(formElement: HTMLFormElement, name: string) {
  const fields: string[] = [];

  for (const element of formElement.elements) {
    if (isInputElement(element) && element.name.startsWith(name)) {
       fields.push(element.name);
    }
  }

  return fields;
}
```

oxc:
Thanks, I will look into this an prepare a new PR.

oxc:
Hi @edmundhung, I looked into your suggestion. It will probably work for all my own use cases, but not all conceivable ones.

More specifically, it will not return all errors that will be returned for this fieldset when validate is called without name.

Especially, it will not report errors on missing values for which there was no field in the DOM, for example. Consider a nested array with its own insert intent button.

Nevertheless, I think it's useful to have this feature to validate multiple specific fields at the same time, so I will prepare a PR for it.

oxc:
Would you think different about _this_ PR if I renamed to new property from `recursive` to `allErrors`? That would make it more in line with how it will get used. If you pass `allErrors: true`, it will include all errors in the output that would be returned for this field that would be returned if you passed no field name. 
In fact, I still think this should be the default behaviour.

edmundhung:
> More specifically, it will not return all errors that will be returned for this fieldset when validate is called without name.

I am not sure if I follow. I think that's what the array of names solves as I expect people to provide all the field names instead. Do you mean the expectation of validating without a name specified doesn't match?

> Especially, it will not report errors on missing values for which there was no field in the DOM, for example. Consider a nested array with its own insert intent button.

This is interesting as it is intentional not to have the newly insert item being validated. Won't it be awkward if you insert a new field and we show the error of that right away? I feel like you have a specific use cases in mind which I might have missed. Can you elaborate more?

oxc:
> > More specifically, it will not return all errors that will be returned for this fieldset when validate is called without name.
> 
> I am not sure if I follow. I think that's what the array of names solves as I expect people to provide all the field names instead. Do you mean the expectation of validating without a name specified doesn't match?




> > More specifically, it will not return all errors that will be returned for this fieldset when validate is called without name.
> 
> I am not sure if I follow. I think that's what the array of names solves as I expect people to provide all the field names instead. Do you mean the expectation of validating without a name specified doesn't match?
> 

Yes, exactly. When I validate a field set (like the Todo in the example), I expect it to show exactly the same errors it will show when I submit the full form. That's currently not possible.

> > Especially, it will not report errors on missing values for which there was no field in the DOM, for example. Consider a nested array with its own insert intent button.
> 
> This is interesting as it is intentional not to have the newly insert item being validated. Won't it be awkward if you insert a new field and we show the error of that right away? I feel like you have a specific use cases in mind which I might have missed. Can you elaborate more?

I might have described that badly. What I meant is: if you have a nested array with insert button, you only have field elements for the single entries, not for the array field itself.

Consider having such an array inside a field set, e.g. a hypothetical "labels" array for each task in the Todo list example.

If you have constraints on the array field ("Add at least one tag"), they won't be shown if you validate the surrounding field set.

They will, however, show once the full form is submitted. This is the discrepancy I would like to solve.

oxc:
I have create a StackBlitz based on the multi-validate PR (#878) that demonstrates the problem, based on your suggestion to collect the field names from the Form element:

https://stackblitz.com/edit/uzog2zz5?file=src%2Ftodos.tsx

1. Open the Todos List example
2. Add a Task
3. Click the Clear button to enable the Validate button*
4. Click Validate button
5. Note that only the Content field error shows up in the Task's error list
6. Click the Save button at the bottom of the Form
7. Note that now also the Error for the Tags array shows up in the Task's error list

* perhaps I did something wrong there, but I guess at the time of first rendering, there are no fields in the Form with the tasks' name prefix, which would be another drawback of your suggestion.

edmundhung:
As commented in https://github.com/edmundhung/conform/pull/878#issuecomment-3166049037, you will be able to validate nested fields in v2.

Let me know if you have any questions 🙌🏼

PR #835, opened 2024-12-13
feat: react 19 support

jacob-ebey:


edmundhung:
Hey, I'm not sure why this specific changes fix React 19 support. The only issue I'm aware of with React 19 is related to the auto form reset behavior with server actions, where it would reset Conform state as well when it happens. 

This is now addressed with the v2 APIs and you can see a working React Router example using React 19 at https://github.com/edmundhung/conform/tree/main/examples/react-router.

As this is already working with React 19, I am gonna close this for now. Thanks.

PR #843, opened 2024-12-26
chore: update tutorial regarding empty string default value support

WowWatchdoge:
After https://github.com/edmundhung/conform/pull/741 , Conform will always strip empty values to `undefined`.
It seems that we also need to remove preprocess() function and comments from tutorial.


edmundhung:
Thanks for the PR, @WowWatchdoge ! I think there might be some confusion here as that part of the tutorial is to show what happens before introducing Conform. As you follow the tutorial, it will ask you changing the schema just like your suggestion.

PR #859, opened 2025-02-10
docs: fix typo

yukiomoto:
Thank you for such a nice library. I really like it.

I fixed some typos that I found in the document and comment.

PR #860, opened 2025-02-10
chore: upgrade corepack in ci

yukiomoto:
I made #859 but the ci failed apparently with this issue ( https://github.com/nodejs/corepack/issues/612 ).

I added `npm i -g corepack@latest` to avoid the error.


edmundhung (approved):
Thank you!

PR #864, opened 2025-02-13
Fix type coercion for branded types

### LINKED ISSUE #863: Type coercion does not work with branded types

Type coercion (like described in https://conform.guide/api/zod/parseWithZod#tips) does not work with branded types.

### Conform version

v1.2.2

```ts
const schema = z.object({ foo: z.string() }).brand("BrandedObject")

const formdata = new FormData()
formdata.append("foo", "") // should be passed as `undefined` to schema

parseWithZod(formData, { schema }) // => expected error: ["Required"], actual: none
```

_No response_

### Screenshots or Videos

_No response_

### Additional context

_No response_

krsilas:
Closes #863

### DIFF
--- a/.changeset/early-buckets-complain.md
+++ b/.changeset/early-buckets-complain.md
@@ -1,0 +1,5 @@
+---
+'@conform-to/zod': patch
+---
+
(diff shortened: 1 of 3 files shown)


edmundhung (approved):
Thank you!

PR #871, opened 2025-02-23
feat(conform-zod): add `disableAutoCoercion` option with `coerceFormValue` helper

edmundhung:
This PR adds the ability to disable the default auto coercion behavior by setting the `disableAutoCoercion` option to `true` in `parseWithZod`.

```ts
function Example() {
    onValidate({ formData }) {
      return parseWithZod(formData, {
        schema,
        disableAutoCoercion: true,
      });
    },
  });

  // ...
}
```

You can then manage how the form value is parsed yourself, or use the new `coerceFormValue` helper to coerce form value, like this:

```ts
import { parseWithZod, unstable_coerceFormValue as coerceFormValue } from '@conform-to/zod';
import { z } from 'zod';

// Coerce the form value with default behaviour
  z.object({
    // ...
  })
);

// Coerce the form value with default coercion overrided
  z.object({
    ref: z.number()
    date: z.date(),
    amount: z.number(),
    confirm: z.boolean(),
  }),
  {
    // Trim the value for all string-based fields
    // e.g. `z.string()`, `z.number()` or `z.boolean()`
    string: (value) => {
      if (typeof value !== 'string') {
         return value;
      }

      const result = value.trim();

      // Treat it as `undefined` if the value is empty
      if (result === '') {
         return undefined;
      }

      return result;
    },

    // Override the default coercion with `z.number()`
    number: (value) => {
      // Pass the value as is if it's not a string
      if (typeof value !== 'string') {
        return value;
      }

      // Trim and remove commas before casting it to number
      return Number(value.trim().replace(/,/g, ''));
    },

    // Disable coercion for `z.boolean()`
    boolean: false,
  },
);
```

This helper is extracted from `parseWithZod` and shares the same coercion logic. It will be released with the `unstable_` prefix in `v1.3.0` and is expected to be marked stable soon. Lock your version to the patch version range (e.g. `~1.3.0`) if you want to use this feature without unexpected changes.

### DIFF
--- a/docs/api/zod/coerceFormValue.md
+++ b/docs/api/zod/coerceFormValue.md
@@ -72,6 +72,23 @@ const schema = coerceFormValue(
   }),
   {
     defaultCoercion: {
+      // Trim the value for all string-based fields
+      // e.g. `z.string()`, `z.number()` or `z.boolean()`
+      string: (value) => {
+        if (typeof value !== 'string') {
+          return value;
+        }
+
+        const result = value.trim();
+
+        // Treat it as `undefined` if the value is empty
+        if (result === '') {
+          return undefined;
+        }
+
+        return result;
+      },
+
       // Override the default coercion with `z.number()`
       number: (value) => {
         // Pass the value as is if it's not a string
@@ -110,7 +127,7 @@ const schema = z.object({
 
 ### Define custom coercion
 
-You can define custom coercion for a specific schema by setting the `defineCoercion` option.
+You can customize coercion for a specific schema by setting the `customize` option.
 
 ```ts
 import {
@@ -132,7 +149,7 @@ const schema = coerceFormValue(
     metadata,
   }),
   {
-    defineCoercion(type) {
+    customize(type) {
       // Customize how the `metadata` field value is coerced
       if (type === metadata) {
         return (value) => {
@@ -145,6 +162,7 @@ const schema = coerceFormValue(
         };
       }
 
+      // Return `null` to keep the default behavior
       return null;
     },
   },
--- a/docs/api/zod/coerceFormValue.md
+++ b/docs/api/zod/coerceFormValue.md
@@ -12,7 +12,7 @@ The following rules will be applied by default:
 2. If the schema is `z.number()`, trim the value and cast it with the `Number` constructor
 3. If the schema is `z.boolean()`, treat the value as `true` if it equals to `on` (Browser default `value` of a checkbox / radio button)
 4. If the schema is `z.date()`, cast the value with the `Date` constructor
-5. If the schema is `z.bigint()`, cast the value with the `BigInt` constructor
+5. If the schema is `z.bigint()`, trim the value and cast it with the `BigInt` constructor
 
 ## Parameters
 
--- a/docs/api/zod/coerceFormValue.md
+++ b/docs/api/zod/coerceFormValue.md
@@ -20,9 +20,13 @@ The following rules will be applied by default:
 
 The zod schema to be enhanced.
 
-### `options.skipCoercion`
+### `options.defaultCoercion`
 
-Optional. You can use it to skip type coercion for a specific schema.
+Optional. Set it if you want to [override the default behavior](#override-default-behavior).
+
+### `options.defineCoercion`
+
+Optional. Use it to [define custom coercion](#define-custom-coercion) for a specific schema.
 
 ## Example
 
@@ -59,25 +63,47 @@ function Example() {
 
 ### Override default behavior
 
-You can override the default coercion logic without disabling the type coercion by casting the value yourself inside `z.preprocess`.
+You can override the default coercion by specifying the `defaultCoercion` mapping in the options.
 
 ```ts
 const schema = coerceFormValue(
   z.object({
-    // Override how the `amount` field value is coerced
-    amount: z.preprocess((value) => {
-      // If no value is provided, return `undefined`
-      if (!value) {
-        return undefined;
-      }
-
-      // Clear the formatting and cast the value to number
-      return Number(value.trim().replace(/,/g, ''));
-    }, z.number()),
-
-    // Keep the default behavior
-    number: z.number(),
+    // ...
   }),
+  {
+    defaultCoercion: {
+      // To trim the value for all string-based fields
+      // e.g. `z.string()`, `z.number()` or `z.boolean()`
+      string: (value) => {
+        if (typeof value !== 'string') {
+          return value;
+        }
+
+        const result = value.trim();
+
+        // Treat it as `undefined` if the value is empty
+        if (result === '') {
+          return undefined;
+        }
+
+        return result;
+      },
+
+      // To override the default coercion with `z.number()`
+      number: (value) => {
+        // Pass the value as is if it's not a string
+        if (typeof value !== 'string') {
+          return value;
+        }
+
+        // Trim and remove commas before casting it to number
+        return Number(value.trim().replace(/,/g, ''));
+      },
+
+      // To disable coercion for `z.boolean()`
+      boolean: false,
+    },
+  },
 );
 ```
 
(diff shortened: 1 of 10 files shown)


edmundhung:
@chimame This is the same as the `coerceZodFormData` helper I show you earlier. I think this would be a good step to start giving people more control on how form value is parsed and towards the direction we are going. 

I renamed it to `zodFormValue` just because `coerceZodFormData` sounds like transforming data instead of enhancing the schema 🤔 Happy to discuss more on the naming.

@jacobparis You mentioned that you were fighting with the coercion logic recently. Would this solve the issue you had?

edmundhung:
>  Since it only converts the schema, how about the name coerceZodSchema?

The idea of `zodFormValue` were based on zod's array api: `z.array(schema)` in which `zodFormValue` will feels like `z.formValue(schema)`. 🤣 

I have a hard time deciding the name. As I feel like we need quite a few keywords in this API: `coerce`, `zod`, `form`.

`coerceFormValueWithZod` might be a good idea but it's really long...

chimame:
Ever since I saw this Pull Request today, I haven't been able to stop thinking about naming. 😂

I thought the name "to validate FormData" was confusing, so I thought adding that would mean it would end up being named `coerceFormValueWithZod` or `coerceFormDataWithZod`. It's long, but it's better than not knowing what it means.

jacobparis:
I don't think "WithZod" is needed in the name since you export it from a zod named package

My vote is coerceFormValue, which I'd use in this way in a custom zod file

```ts
// This file allows creating custom zod helpers
import * as z from 'zod';
export * from 'zod';

/**
 * @example z.coerceFormValue(z.number())
 */
export { coerceFormValue } from '@conform-to/zod';

/**
 * Coerces a checkbox value to a boolean
 *
 * @example z.checkbox()
 */
export function checkbox() {
	return z.preprocess(
		(value) => (value === 'on' ? true : value === 'off' ? false : undefined),
		z.boolean().optional(),
	);
}

/**
 * Radix selects don't allow empty values for options, so we need a magic keyword to represent null
 *
 * @example z.optionValue()
 */
export function optionValue() {
	return z.string().transform((value) => (value === 'NULL' ? null : value));
}

export function email() {
	return z.string().transform((value) => value.toLowerCase());
}
```

edmundhung:
> I don't think "WithZod" is needed in the name since you export it from a zod named package

Aren't you were saying `parse` being overloaded by many packages before `parseWithZod` is introduced 🤔 That was the main reason why I named most APIs with the `zod` name included.

I know the pattern you mentioned is suggested by zod but I don't think many people follows.

jacobparis:
I can agree with parse itself being overloaded, I get about 20 import suggestions on that one alone, but I don't think that needs to affect every other export. Coerce isn't a very common function name on its own, and especially scoping it down by form data should be uniquely identifable enough

edmundhung:
> Coerce isn't a very common function name on its own, and especially scoping it down by form data should be uniquely identifable enough

Fair enough. Then the question now become whether I should rename `getZodConstraint` to `getConstraint` for consistency... just kidding.

I am fine with either `coerceFormValue` or `coerceFormValueWithZod`. @chimame any preference?

chimame:
It's a difficult choice, but I want to choose `coerceFormValue`.

I could have named it `coerceFormValueWithZod`, but it's rare to use both zod and valibot schema libraries in an application. So I thought it might be better not to add `WithZod` to the API name.

edmundhung:
@chimame Sorry for the delay. I have updated the options and it is ready to merge anytime. I don't have the docs translated to Japanese but I think it's fine to be in english for now.

packages/conform-zod/coercion.ts
```diff
@@ -73,12 +53,61 @@ export function coerceFile(file: unknown) {
 	return file;
 }
 
+function coerceNumber(value: unknown) {
+	if (typeof value !== 'string') {
+		return value;
+	}
+
+	return value.trim() === '' ? value : Number(value);
```
chimame, line 61: Is there a reason why `Number` and `BigInt` are not `undefined` for empty strings?

chimame (commented):
Thank you for contacting me.

I also changed the API in `@conform-to/valibot` based on this Pull Request.
I will translate the documentation into Japanese at a later date.

packages/conform-zod/coercion.ts
edmundhung: That is handled by the string coercion, which is called before number / boolean coercion. I know this is a bit implicit and it would be simpler to not having the concept of composing coercion. But it is a bit tricky to handle optional cases where we don't know the type upfront and that's why we are coercing those value with string / file by default. Composing them makes it more consistent in terms of behavior... I think 🤔

edmundhung: E.g. if you override the default string coercion logic to trim the spaces and you have an optional number schema. Your number schema will always see the string trimmed because we applied the string coercion in the optional schema.

chimame: I see, I understand.
Since `composeCoercion` is used to first perform coercion of the `coerceString` string, an empty string is defined as `undefined`, so there should be no problem. 👍

chimame:
@edmundhung 
Thanks for creating the new API!
This should enable finer control over the validation schema. 🎉

docs/api/zod/coerceFormValue.md
```diff
@@ -162,7 +145,6 @@ const schema = coerceFormValue(
         };
       }
 
-      // Return `null` to keep the default behavior
```
chimame: I thought it would be fine to leave this comment, but should I delete it?

Of course, returning `null` will result in the behavior defined by the `defaultCoercion` option or the default behavior, so would it be easier to understand if you made that clear?

chimame (commented):
I understand.
So you're saying that only `stripEmptyString` is defined as private, and the result of executing `stripEmptyString` will use the behavior of `string()` or `number()` defined in `defaultCoercion` etc.

I think the valibot package should also follow this behavior.

docs/api/zod/coerceFormValue.md
```diff
@@ -0,0 +1,152 @@
...
+const enhancedSchema = coerceFormValue(schema, options);
+```
+
+The following rules will be applied by default:
+
+1. If the value is an empty string / file, pass `undefined` to the schema
+2. If the schema is `z.number()`, trim the value and cast it with the `Number` constructor
+3. If the schema is `z.boolean()`, treat the value as `true` if it equals to `on` (Browser default `value` of a checkbox / radio button)
+4. If the schema is `z.date()`, cast the value with the `Date` constructor
+5. If the schema is `z.bigint()`, cast the value with the `BigInt` constructor
```
chimame, line 15: `trim` is used in the same way as `Number`, so would this be a good explanation?

```suggestion
5. If the schema is `z.bigint()`, trim the value and cast it with the `BigInt` constructor
```

edmundhung:
> I understand. So you're saying that only `stripEmptyString` is defined as private, and the result of executing `stripEmptyString` will use the behavior of `string()` or `number()` defined in `defaultCoercion` etc.
> 
> I think the valibot package should also follow this behavior.

Honestly I still have mixed feeling with the current solution. I like that we simplify the `defaultCoercion` options. But now I am not sure if there is any good use cases to override the string coercion logic as  my original example that trims the value doesn't work well with this design:

Let's say you have a field defined as `z.string().optional()`, we will now strip empty string in the optional schema and the string schema will receive a string with **spaces**. If you try to trim the value with `options.defaultCoercion.string`, you will now end up with a "required" error 😅

edmundhung:
It might be different in Valibot. But my understanding with Zod is that the value goes through the zod schema like this:

`(value) -> optional schema -> (if not undefined) -> string schema`

The current setup will strip empty string in the optional schema, so if the value is `""`, this happens:

`(value) -> (stripEmptyString) -> optional schema -> (optional, skip the wrapped schema)`

But if the value is a string with spaces (e.g. `"  "`) and you try to trim the string and returns `undefined` if it becomes an empty string in `options.coercion.string`:

`(value) -> (stripEmptyString) -> optional schema -> (not undefined, so pass it to the wrapped schema) -> (string coercion occurs, the value becomes undefined) -> string schema -> (required error)`

chimame:
Thank you for the detailed explanation. I understand.

It's true that this test code passes on the valibot side.
```ts
describe('test', () => {
  test('trim string', () => {
    const originalSchema = object({
      name: optional(string()),
    });

    const schema = coerceFormValue(originalSchema, {
      defaultCoercion: {
        string: (value) => {
          if (typeof value !== 'string') {
            return value;
          }
          return value.trim() === '' ? undefined : value.trim();
        }
      }
    })
    const result = safeParse(schema , { name: '  ' });

    expect(getResult(result)).toEqual({
      success: true,
      data: {
        name: undefined,
      },
    });
  });
});
```

This is because the valibot side also processes in the order `optional schema -> string schema`, but the transform, that is, the combination process of `stripEmptyString` and `defaultCoercion`, is controlled so that it is performed only once before `optional schema`. (This makes the recursive process a little more complicated.)

In other words, it is not the valibot's schema check specifications that are the problem, but rather the schema conversion process in `coerceFormValue` that is slightly devised.

It's true that you'll need to change the logic to pass this. 😅

chimame:
I will proceed with adding the `coerceFormValue` API as it is, and the problem of the timing of this forced conversion process can only be solved by modifying the schema conversion logic, so why don't you tackle it as a separate issue later? If possible, I will try to fix it at a later date.

chimame:
The option name change has also been applied to valibot.

There is absolutely no problem with such changes. Just be careful when creating new names. Also, your top priority right now should be enjoying your trip😉. Library maintenance can come later.

PR #876, opened 2025-03-09
feat(conform-valibot): add support valibot

chimame:
## Overview

Add a package that supports valibot based on [`conform-to-valibot`](https://github.com/chimame/conform-to-valibot).

`@conform-to/valibot` extends the schema so that FormData can validate the valibot schema defined by the user.

```ts
import { object, string } from "valibot";

const schema = object({
  name: string()
})

// Convert to a schema for FormData validation
const validationSchema = coerceFormValue(schema)
// object({
//   name: pipe(
//     unknown(),
//     transform(value => value === '' ? undefined : value),
//     string()
//   )
// })
```

There is a low level API being added to `@conform-to/zod`. `@conform-to/valibot` also adds the same API. https://github.com/edmundhung/conform/pull/871

## Changes
- Create based on the package structure of `@conform-to/zod`
  - Test files placed in `packages/conform-valibot/tests/`
  - Changed build files to output to `packages/conform-valibot/dist/`
- Add `coerceFormValue` low-level API to convert the valibot schema to validate FormData.
  - Add `disableAutoCoercion` option to `parseWithValibot` API to disable schema conversion by `coerceFormValue`

## TODO

The following additional steps are required.

- [x] Create documentation
- [x] Add a documentation link to the sidebar menu on the Conform website
- [x] Create the `coerceFormValue` API
- [x] Add a valibot schema supported by `getValibotConstraint` (e.g. `objectWithRest`)

### DIFF
--- a/packages/conform-valibot/coercion.ts
+++ b/packages/conform-valibot/coercion.ts
@@ -92,7 +92,7 @@ function coerceNumber(value: unknown) {
 		return value;
 	}
 
-	return value.trim() === '' ? undefined : Number(value);
+	return value === '' ? undefined : Number(value);
 }
 
 /**
@@ -133,7 +133,7 @@ function coerceBigInt(value: unknown) {
 		return value;
 	}
 
-	if (value.trim() === '') {
+	if (value === '') {
 		return undefined;
 	}
 
--- a/packages/conform-valibot/coercion.ts
+++ b/packages/conform-valibot/coercion.ts
@@ -51,10 +51,12 @@ function coerce<T extends GenericSchema | GenericSchemaAsync>(
 }
 
 /**
- * Helpers for coercing string value
- * Modify the value only if it's a string, otherwise return the value as-is
+ * Strip empty strings and convert them to undefined.
+ *
+ * @param value The value to be checked and coerced.
+ * @returns The coerced value or undefined if the input is an empty string.
  */
-function coerceString(value: unknown) {
+function stripEmptyString(value: unknown) {
 	if (typeof value !== 'string') {
 		return value;
 	}
@@ -67,10 +69,12 @@ function coerceString(value: unknown) {
 }
 
 /**
- * Helpers for coercing file
- * Modify the value only if it's a file, otherwise return the value as-is
+ * Strip empty files and convert them to undefined.
+ *
+ * @param file The file to be checked and coerced.
+ * @returns The coerced value or undefined if the input is an empty file.
  */
-function coerceFile(file: unknown) {
+function stripEmptyFile(file: unknown) {
 	if (
 		typeof File !== 'undefined' &&
 		file instanceof File &&
@@ -92,7 +96,7 @@ function coerceNumber(value: unknown) {
 		return value;
 	}
 
-	return value === '' ? undefined : Number(value);
+	return value.trim() === '' ? value : Number(value);
 }
 
 /**
@@ -133,8 +137,8 @@ function coerceBigInt(value: unknown) {
 		return value;
 	}
 
-	if (value === '') {
-		return undefined;
+	if (value.trim() === '') {
+		return value;
 	}
 
 	try {
@@ -158,7 +162,7 @@ function coerceArray<T extends GenericSchema | GenericSchemaAsync>(type: T) {
 
 		if (
 			typeof output === 'undefined' ||
-			typeof coerceFile(coerceString(output)) === 'undefined'
+			typeof stripEmptyFile(stripEmptyString(output)) === 'undefined'
 		) {
 			return [];
 		}
@@ -171,6 +175,16 @@ function coerceArray<T extends GenericSchema | GenericSchemaAsync>(type: T) {
 	}
 
 	return pipe(unknown, vTransform(transformFunction), type);
+}
+
+/**
+ * Compose two coercion functions
+ * @param a The first coercion function
+ * @param b The second coercion function
+ * @returns The composed coercion function
+ */
+function compose(a: CoercionFunction, b: CoercionFunction): CoercionFunction {
+	return (value) => b(a(value));
 }
 
 /**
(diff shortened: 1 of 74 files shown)


chimame:
@edmundhung 

I created a package that supports valibot while providing the same API interface as #871.

There was one thing that bothered me.
The definition of `string` passed to the `defaultCoercion` option of `coerceFormValue`. In #871, if you specify `string` as an option, it is common to processes such as `number` and `date`, and conversion of empty strings etc. is done by this `string` definition.
Since the user will be using `string`, won't they mistakenly think that it is only a definition related to string fields? Therefore, I tried to make the `string` option only apply to schemas related to string fields (it does not affect `number` or `date`).

I would like your opinions on this.

packages/conform-valibot/coercion.ts
```diff
@@ -0,0 +1,516 @@
...
+			};
+
+			return {
+				transformAction: undefined,
+				schema: objectWithRestSchema,
+			};
+		}
+	}
+
+	return coerce(type, options.defaultCoercion.string);
```
edmundhung, line 435: I am curious why you defaults to string coercion here as you have already handled the string case above :thinking

edmundhung:
> Since the user will be using `string`, won't they mistakenly think that it is only a definition related to string fields? Therefore, I tried to make the `string` option only apply to schemas related to string fields (it does not affect `number` or `date`).

Yes. Ideally it should only apply to one type. The only reason why I have to compose them is because of consistency with optional schema (e.g. `z.number().optional()`) where I need to strip the empty string in the optional schema to make sure zod doesn't validate it with the number schema if it is an empty string.

But I just have a quick look of your implementation and it looks like you don't have to coerce the value specifically for this case except for the array coercion. Is valibot handles optional schema in a different way?

One potential solution might be to separate the logic that strip empty string / file out of the coercion. Then there is no need to apply any custom string coercion logic in the optional schema as we can continue stripping the empty value as it was and apply the string coercion logic only in the base schema (e.g. `z.string()` / `z.number()` ...) 🤔

edmundhung (commented):
Thanks so much for working on this. This is looking really good!

I am wondering if we should release this following the conform version, basically bumping all version together whenever any of the packages is change. The problem I see here is that valibot v1 seems to be very close to be landed. If we need to update the minimum valibot version in the peer dependency, this will be a breaking change and forcing the rest of the package to v2 😅 

We will have the same issue when Zod v4 is released. So maybe it's time to start releasing each package seperately?

packages/conform-valibot/package.json
```diff
@@ -0,0 +1,66 @@
...
+		"directory": "packages/conform-zod"
+	},
+	"bugs": {
+		"url": "https://github.com/edmundhung/conform/issues"
+	},
+	"dependencies": {
+		"@conform-to/dom": "workspace:*"
+	},
+	"peerDependencies": {
+		"valibot": ">= 0.32.0"
```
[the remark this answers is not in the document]
chimame: Valibot 0.32 has some breaking changes internally (how schema data is stored internally in valibot) and externally (schema validation behavior, i.e. `safeParse` behavior). Also, 1.0 has some breaking changes externally (schema validation behavior, i.e. `safeParse` behavior), but no major breaking changes internally.

So, since `@conform-to/valibot` affects more internal changes, I set the minimum version to 0.32.

However, since it is a new package, it may be better to support it from 1.0 (which is of course currently an RC).

packages/conform-valibot/coercion.ts
chimame: This is a `switch`, that is, to make it the return value of the `enableTypeCoercion` function when it does not match the schema definition of a supported valibot.

Basically, this should be the schema definition of a supported valibot, but in case the number of schemas that can be defined in valibot increases or a schema that is not currently supported is passed, it is no exaggeration to say that this is forcibly set as the return value of the `enableTypeCoercion` function.

packages/conform-valibot/package.json
```diff
@@ -0,0 +1,66 @@
...
+		"dev:js": "pnpm run build:js --watch",
+		"dev:ts": "pnpm run build:ts --watch",
+		"dev": "pnpm run \"/^dev:.*/\"",
+		"test": "vitest",
+		"prepare": "pnpm run build"
+	},
+	"repository": {
+		"type": "git",
+		"url": "https://github.com/edmundhung/conform",
+		"directory": "packages/conform-zod"
```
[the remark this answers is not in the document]
chimame: It's a simple mistake. Thanks for finding it.

chimame:
If possible, I would like to keep the behavior of `string()` defined in `defaultCoercion` as it is now, without affecting `number()`. Although the specifications will be different from zod, I would like to consider the behavior if I receive any feedback from users.

> I am wondering if we should release this following the conform version, basically bumping all version together whenever any of the packages is change. 

It is certainly a difficult problem.
I think that there will be many cases where changes occur to `@conform-to/zod` and `@conform-to/valibot`, but not to `@conform-to/dom`, so it would be preferable to release each package separately.

packages/conform-valibot/coercion.ts
chimame: If you think about it rationally, you're right, there's no need for conversion, so I've made it so that the value is returned as is.

edmundhung:
Let me try separating the logic that strip empty string / file out of the zod coercion tonight. If people override the string coercion and stop it from striping empty value, they should just use their own solution instead of this helper.

> it would be preferable to release each package separately.

👍🏼 Let's make this `v1.0.0`. 

> it may be better to support it from 1.0

There's no need to restrict the new package to v1 if we can support both version. If we are confident that the internals won't change in v1, I think we can ship it as is. Do you know the timeline for the valibot v1 release? Is it expected to release in days, weeks or months?

chimame:
The 1.0.0 release of valibot is scheduled for next week. 👍

edmundhung:
I will try to publish this as v0.0.1 tonight so we can test it out first.

edmundhung:
v0.0.1 is out: https://www.npmjs.com/package/@conform-to/valibot 🚀 

The pre-release CI should work now.

edmundhung:
I took another look of your implementation as I was curious how you manage to handle the coercion. My understanding is that you extract the coercion logic from the wrapped schema and apply it on top of the optional schema which is really nice! But I notice there are some edge cases this doesn't work, like this one:

```ts
test('optional + union', () => {
  const schema = object({
    a: optional(union([literal('a'), literal('b'), literal('c')])),
  });

  expect(parseWithValibot(createFormData('a', ''), { schema })).toMatchObject({
    status: 'error',
    error: {},
  });
});
```

I would expect this to be successful but the union inside the optional schema makes it impossible to extract the coercion logic. To be fair, this is quite rare and it might be an acceptable trade-offs. Just sharing in case you are not aware of it 👍🏼

chimame:
> My understanding is that you extract the coercion logic from the wrapped schema and apply it on top of the optional schema which is really nice!

Yes, Valibot is extracting the transformation logic from the wrapped schema to validate and running the coercion logic first.

> I would expect this to be successful but the union inside the optional schema makes it impossible to extract the coercion logic.

It is true that an error occurs in this case. This is because the `union` schema is wrapped in an `optional` schema, but there are factors that make it impossible to determine a single coercion schema for the `union`. For example, since it is a `union`, if not only `literal` but also `number` is allowed, a single coercion logic cannot be determined. I think this is the only way to validate such cases.

```ts
test('optional + union', () => {
  const schema = object({
    a: union([optional(literal('a')), number()]),
  });
  
  expect(parseWithValibot(createFormData('a', ''), { schema })).toMatchObject({
    status: 'success',
    value: { a: undefined },
  });
});
```

Thank you for letting me know the pattern that causes the error. It's a bit tricky, but I have a workaround, so I'll leave it as it is. 🙏

packages/conform-valibot/package.json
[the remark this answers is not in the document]
chimame: Just to be sure, I reverted the version back to 0.32 and tested it, and confirmed that it ran without any problems.

chimame:
I came up with a way to support the `optional(union())` schema, so I fixed it. It should now be easier to use. 👍

chimame:
All the documentation has been revised so it can be merged at any time.

packages/conform-valibot/package.json
```diff
@@ -0,0 +1,66 @@
...
+	"repository": {
+		"type": "git",
+		"url": "https://github.com/edmundhung/conform",
+		"directory": "packages/conform-valibot"
+	},
+	"bugs": {
+		"url": "https://github.com/edmundhung/conform/issues"
+	},
+	"dependencies": {
+		"@conform-to/dom": "workspace:*"
```
edmundhung, line 41: I think we will need to use a [Caret Ranges](https://docs.npmjs.com/cli/v6/using-npm/semver#caret-ranges-123-025-004) here as we will not be bumping the valibot package version together with `@conform-to/dom` everytime. The caret range should allow people to use a newer version of `@conform-to/dom` without issue:

```suggestion
		"@conform-to/dom": "workspace:^"
```

edmundhung (commented):
Sorry for the wait! This is looking good. I will give it a few more tests tonight but I think we can release it tomorrow 👍🏼

edmundhung (approved):
LGTM!

PR #877, opened 2025-03-11
Update tutorial.md

AlexWebLab:
the property of the returned object is fieldErrors according to the previous examples

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -157,12 +157,12 @@ export default function ContactUs() {
           name="email"
           defaultValue={result?.payload.email}
           required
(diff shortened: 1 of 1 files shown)


edmundhung (approved):
Thank you! Mind updating the ja tutorial docs as well? We need to keep them consistent as much as we can. 🙏🏼

AlexWebLab:
> Thank you! Mind updating the ja tutorial docs as well? We need to keep them consistent as much as we can. 🙏🏼

here we go: [https://github.com/edmundhung/conform/pull/880](https://github.com/edmundhung/conform/pull/880)

PR #878, opened 2025-03-12
feat: allow multiple fields to be validated in one intent

oxc:
As suggested by @edmundhung in https://github.com/edmundhung/conform/issues/830#issuecomment-2676301855

### DIFF
--- a/playground/app/routes/form-control.tsx
+++ b/playground/app/routes/form-control.tsx
@@ -75,6 +75,14 @@ export default function FormControl() {
 						</button>
 						<button
 							className="rounded-md border p-2 hover:border-black"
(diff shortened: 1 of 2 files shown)


edmundhung (commented):
This looks good to me. Can you help adding an integration test for it? You can update the form-control test with a new button that validate multiple fields and assert if it works properly. Thanks for working on this!

oxc:
I've added a test for this case in form-control as you suggested.

oxc:
Is anything else required to get this merged?

oxc:
@edmundhung, is this in any way obsolete or no longer useless? Or is there another reason this cannot get merged?

PR #879, opened 2025-03-12
Return all errors for fields if requested in validate intent

### LINKED ISSUE #828: Cannot validate nested object fields

oxc:
Follow-up on https://github.com/edmundhung/conform/pull/830#issuecomment-2716353327, based on #878 

Fixes #828

### DIFF
--- a/.changeset/great-roses-smoke.md
+++ b/.changeset/great-roses-smoke.md
@@ -1,0 +1,5 @@
+---
+'@conform-to/dom': minor
+---
+
+Return all errors for fields if requested in validate intent
--- a/.changeset/stupid-needles-rhyme.md
+++ b/.changeset/stupid-needles-rhyme.md
@@ -1,0 +1,5 @@
+---
+'@conform-to/dom': minor
+---
+
+Allow validating multiple fields in one validate intent
--- a/packages/conform-dom/form.ts
+++ b/packages/conform-dom/form.ts
@@ -332,7 +332,13 @@ function handleIntent<Error>(
 	switch (intent.type) {
 		case 'validate': {
 			if (intent.payload.name) {
-				meta.validated[intent.payload.name] = true;
+				if (typeof intent.payload.name === 'string') {
+					meta.validated[intent.payload.name] = true;
+				} else {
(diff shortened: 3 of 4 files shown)


oxc:
Here is a StackBlitz that demonstrates the difference of the Intent with and without the new field: https://stackblitz.com/edit/4t2buwcu?file=src%2Ftodos.tsx,src%2Findex.css

edmundhung:
Thanks for working on this! I will take a look over the weekend.

edmundhung (commented):
I find it a bit weird to mark all fields under a prefix validated the more I think about this. As we generally avoid showing errors of a field until the user interact with it, it will be strange if we mark a field list as validated and the user starting seeing errors from fields just inserted on the form before they have a chance filling in anything. This is why I feel like accepting an array of names is more appropriate as we could restrict it to only fields that are visible to users.

PR #887, opened 2025-03-21
docs: the Return value of `useForm` Function

ogadra:
Some documents state that `form` is the same object as the one returned by the `useForm` hook. However, there is no dedicated description of the object returned by the `useForm` hook.

This PR aims to compensate for this issue.

Please let me know if there are any errors or omissions.


chimame (approved):
Thank you for adding the documentation. Let's merge it.

lhapaipai:
Hi @ogadra , thank you very much for this extra documentation !
I wonder what is the difference between.
`defaultValue` that we can pass to the hook `useForm`. and `initialValue` that the hook return. why a different name ? what is the difference between the two ?

ogadra:
Hi, @lhapaipai . I'm really sorry for the delayed response. Thank you for your patience.

I'm not fully confident, but regarding your question about defaultValue and initialValue:

- `defaultValue` is the value explicitly provided by the programmer when calling the `useForm` hook. It represents the initial form state defined in your code and is fully under your control.
  - Controlled by the programmer, unless explicitly changed.
- `initialValue`, on the other hand, is the value returned by the `useForm` hook, which can sometimes differ from your provided `defaultValue`. Specifically, if a form submission results in errors, the library internally uses the result from the failed submission (`lastResult`) to overwrite the `initialValue`. This means `initialValue` might reflect the user's previously submitted data rather than the original default. 
  - Managed internally; may be dynamically updated based on user submissions and errors.

You can find an example of this logic here:
https://github.com/edmundhung/conform/blob/main/packages/conform-dom/form.ts#L269

lhapaipai:
hi @ogadra ,
Thank you very much for your answer !
I also noted that it is `initialValue` which is used when calling the reset method.
have a good day

PR #902, opened 2025-04-12
feat(conform-zod): add support zod v4

chimame:
## Overview

https://github.com/colinhacks/zod/pull/4074

Zod v4 is scheduled for release, and `@conform-zod` provides a new `@conform-to/zod/v4` module to support it.

## Details

Our policy for supporting Zod v4 is as follows:

- We will support both `zod` and `zod/v4`, `zod/v4-mini` in Zod v4.
- ~~Zod libraries that modules that support Zod v4 depend on will be changed to `@zod/core` and `@zod/mini`.~~
- We will continue to support Zod v3.

The new Zod v4 API only depends on zod/v4-mini, so users should see smaller bundle sizes when using Zod v4 API.

## Changes

- Created a new module in `@conform-to/zod/v4` to support Zod v4.
- Changed the location of `tests/conform-zod.spec.ts` to split test files by schema.
- Changed the output destination of the build file of `@conform-to/zod` to `dist/` due to the change in the location of the test code.

`packages/conform-zod/v4/coercion.ts` and `packages/conform-zod/coercion.ts` have similar content. However, we have placed them here because we believe that it is less maintenance cost to maintain them duplicated than to end support for Zod v3 in the future or to make the code more complicated by standardizing the processing for v3 and v4.


chimame:
I think this is the reason why safeParseAsync in Zod v4 does not work properly.

https://github.com/colinhacks/zod/pull/4132

colinhacks:
I've only skimmed this but it's looking quite good. Ideally libraries like conform will depend only on `@zod/core`. Both `zod` and `@zod/mini` depend on it — `zod` does not depend on `@zod/mini`. There are some docs it here:

chimame:
@colinhacks

Thank you for your advice.
If we can make it compatible with both Zod v3 and v4, it will be a user-friendly update, so this information is very much appreciated.

chimame:
With these changes, it now works with Zod v4.

chimame:
@edmundhung 

You don't need to review the whole thing yet.
Is it okay to split the test files and change their location like in this commit?

https://github.com/edmundhung/conform/pull/902/commits/d85de2c765f595daeb028ac87e3fab82132a151d

edmundhung:
> Is it okay to split the test files and change their location like in this commit?

I don't have a strong preference on this. Feel free to split 👍🏼

edmundhung:
Thanks for looking into this! I think it is quite important to keep v3 support at this stage as people often keep using the old version for a fairly long time.

A few ideas:

1. Exports a new set of APIs (e.g. `parseWithZod4`) that indicate zod 4 support 😅 
2. Ships a new package `@conform-to/zod4` 😓 
3. Setup module exports, e.g. `@conform-to/zod/v4` 🤔

Personally, I think (3) might be the best compromise here. We will have to maintain the integration for both v3 and v4 but v3 is quite stable and so the maintenance cost should be minimal.

chimame:
Thank you for the great idea.
I also think that `@conform-to/zod/v4` is the most user-friendly format. I also think that this is possible, so I will work on it.

packages/conform-zod/v4/coercion.ts
```diff
@@ -0,0 +1,415 @@
...
+	// Return the cached schema if it's already processed
+	// This is to prevent infinite recursion caused by z.lazy()
+	if (result) {
+		return result;
+	}
+
+	let schema: $ZodType = type;
+	const zod = (type as unknown as $ZodTypes)._zod;
+	const def = zod.def;
+	const constr = zod.constr;
```
chimame, line 169: The original schema defined by the user inside Zod is stored in `constrt`. `transform` is used to redefine the original schema by applying schema operations to the original schema.

chimame:
@edmundhung 

I've created a module that works with Zod v4.
I'm in no hurry, so please do a code review.

If you were to release this, would you plan to do so after Zod v4 is released?

packages/conform-dom/formdata.ts
```diff
@@ -64,7 +64,7 @@ export function getPaths(name: string | undefined): Array<string | number> {
  * const name = formatPaths(['todos', 0, 'content']); // "todos[0].content"
  * ```
  */
-export function formatPaths(paths: Array<string | number>): string {
+export function formatPaths(paths: Array<PropertyKey>): string {
```
edmundhung, line 67: I am not sure if we could support `symbol` here. We might be able to serialize it in some form but we will not be able restore it back as a symbol. 

We might have to throw an error with zod 4 if the paths contain any symbol.

packages/conform-zod/v4/coercion.ts
```diff
@@ -0,0 +1,415 @@
+import type { $ZodType, $ZodTypes } from '@zod/core';
+import { lazy, pipe, transform, union } from '@zod/mini';
```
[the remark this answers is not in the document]
edmundhung: I wonder if we can do some tricks like this:

```ts
let z: typeof import('@zod/mini') | undefined;

try {
  z = require('@zod/mini');
} catch (e) {
  // If @zod/mini is not available, look for zod
  z = require('zod');
}

// Assuming @zod/mini and zod exports these APIs in the same name
// Now we can do...
z.lazy()
z.pipe()
```

packages/conform-zod/v4/coercion.ts
```diff
@@ -0,0 +1,415 @@
...
+		//         type: z.literal('a'),
+		//         number: z.number(),
+		//       })
+		//     ),
+		// ]);
+		// The `options` of `discriminatedUnion` basically defines the `object` schema, but `@conform-to/zod` requires `transform` to convert Object to an empty object when it is `undefined`, so the `union` schema is used.
+		// `discriminatedUnion` and `union` are strictly different schemas, but they behave similarly.
+		schema = union(
+			def.options.map((option) => enableTypeCoercion(option, options)),
+		);
```
edmundhung, line 262: The example looks like zod 3? Is it still a problem now that we have v4 exported on a different module?

packages/conform-zod/v4/coercion.ts
```diff
@@ -0,0 +1,415 @@
...
+	return schema;
+}
+
+/**
+ * A helper that enhance the zod schema to strip empty value and coerce form value to the expected type with option to customize type coercion.
+ * @example
+ *
+ * ```tsx
+ * import { parseWithZod, unstable_coerceFormValue as coerceFormValue } from '@conform-to/zod';
+ * import { z } from 'zod-4';
```
edmundhung, line 299:
```suggestion
 * import { z } from 'zod';
```

edmundhung (commented):
Thanks so much for working on this 🙌🏼 

I have left a few comments but the changes looks good to me overall! A few additional thoughts:

- It's great that we are testing both zod 4 and zod mini! But there are quite a bit of overlaps on the tests right now. Would it worth merging the tests? Perhaps by creating two set of schema using each package and run them through the same sets of test so we can also make sure they behaves the same. It's fine to keep the v3 tests separated.
- Nitpick: I think we might have to keep v3 around for a relatively long term even when Conform reach v2 and we could just change the default to v4 and expose v3 as `@conform-to/zod/v3`. Should we move the current implementation in a v3 folder similar to v4? 
- We should start thinking about how to document this. It might not be obvious that they will need to use a different imports for zod 4 support.
- It might be better to wait for the zod 4 stable release. As our integration is pretty involved and the users might find it hard to figure out whether it is a bug on Conform or Zod sides.

chimame:
Thank you for your feedback despite the many code changes.

> It's great that we are testing both zod 4 and zod mini! But there are quite a bit of overlaps on the tests right now. Would it worth merging the tests? Perhaps by creating two set of schema using each package and run them through the same sets of test so we can also make sure they behaves the same. It's fine to keep the v3 tests separated.

I separated `zod` and `@zod/mini` because I was concerned that the file might become too large if I managed the test code in a single file. However, I understand your point of view. Let's integrate it as you suggested.

> Nitpick: I think we might have to keep v3 around for a relatively long term even when Conform reach v2 and we could just change the default to v4 and expose v3 as `@conform-to/zod/v3`. Should we move the current implementation in a v3 folder similar to v4?

I think it's a little early to use `@conform-to/zod/v3`. Zod v4 has many changes, and I think it will take some time for users to adapt. If that happens, many users will remain with Zod v3. This would require users of Conform to also change to `@conform-to/zod/v3`, so why not leave it as `@conform-to/zod` for now, but move the modules to the `v3` directory as well as `@conform-to/zod/v4`, so that when Zod v4 becomes the default, you only need to change the `exports` definition in `package.json`?

> We should start thinking about how to document this. It might not be obvious that they will need to use a different imports for zod 4 support.

I was wondering where to add documentation. I think I would add "installation" to the `Overview` of `Getting Started` in the conform.guilde documentation, or add `Installation` to the submenu of `Getting Started` (between `Overview` and `Tutorial`) and describe it. It would be better to describe installation for Zod v3 and v4 as well as valibot and yup.

> It might be better to wait for the zod 4 stable release. As our integration is pretty involved and the users might find it hard to figure out whether it is a bug on Conform or Zod sides.

It's difficult to decide on a release date, but I think it would be fine to release it at the same time as Zod v4. Supporting Zod v4 is difficult, but `@zod/mini` is a welcome change that reduces the bundle size. Some users will upgrade early for that reason. If Conform is supported at that time, that would be great too.

edmundhung:
> I was concerned that the file might become too large if I managed the test code in a single file.

Just to make sure there is no misunderstanding: I am not suggesting merging all tests back to one file. It's just about moving the the individual tests for zod mini schema to the zod 4 tests as they are almost identical 

> but move the modules to the v3 directory
👍🏼

> It would be better to describe installation for Zod v3 and v4 as well as valibot and yup.

👍🏼

>  I think it would be fine to release it at the same time as Zod v4

Let's aim for that! 👏🏼

chimame:
The following issues remain unresolved:

- The behavior of `file` differs between commonjs and esm. This is probably a zod issue.
- Updating `@zod/core` to version 0.8.1 breaks types.

```ts
const fileEsm = file();
const fileCjs = require('@zod/mini').file();
console.log(fileEsm.safeParse('').error, fileCjs.safeParse('').error);
// $ZodError {
//   issues: [
//     {
//       expected: 'file',
//       code: 'invalid_type',
//       path: [],
//       message: 'Invalid input: expected file, received string'
//     }
//   ]
// } $ZodError {
//   issues: [
//     {
//       expected: 'file',
//       code: 'invalid_type',
//       path: [],
//       message: 'Invalid input' <-- different error messages are output.
//     }
//   ]
// }
```

packages/conform-zod/v4/coercion.ts
chimame: This is a type error when defined in Zod v4. If you ignore the type and run it, an exception will be thrown and it will not run correctly.
image

Zod v3 is also slightly different, but the type is an error.
image

It may be possible to avoid this by instantiating `ZodDiscriminatedUnion` as in Zod v3, but I haven't looked into it that far. My guess is that there may not be such a workaround in `@zod/mini`.

chimame:
I asked Zod about these two issues.

- https://github.com/colinhacks/zod/issues/4262
- https://github.com/colinhacks/zod/issues/4263

colinhacks:
Apologies for the inconvenience - I've decided to take a different approach with respect to compatibility and versioning. It should make things a lot easier on your end (but will require some refactoring on this PR I'm afraid): https://github.com/colinhacks/zod/issues/4371

chimame:
@colinhacks 

Good luck with the release of Zod v4.
Also, thank you for taking the time to comment here. 😀

We'll work on this as well to make it compatible with Zod v4.
I have created an issue about the behavior of `discriminatedUnion`, so please let me know if there is any progress.

chimame:
An error occurs in the test of the catch schema, but I have created a pull request for zod.

https://github.com/colinhacks/zod/pull/4382

packages/conform-zod/v4/coercion.ts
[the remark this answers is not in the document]
chimame: Zod v4 will be released with Zod 3.25, so this action is no longer necessary.

packages/conform-dom/formdata.ts
chimame: fixed

packages/conform-zod/v4/coercion.ts
chimame: Zod v4 has been modified to redefine `discriminatedUnion`. The logic has been modified to redefine it from the user-defined schema stored in Zod.

packages/conform-zod/v4/coercion.ts
chimame: fixed

edmundhung (approved):
Love it! Just one minor comment. Let's get this out once zod 3.25 is released 🎉 

Just make sure we updated the example zod version to 3.25 before merging it.

packages/conform-zod/v4/coercion.ts
```diff
@@ -257,24 +257,45 @@ export function enableTypeCoercion<Schema extends $ZodType>(
...
+					const object = new item._zod.constr({
+						...objectDef,
+						shape: Object.fromEntries(
+							Object.entries(objectDef.shape).map(([key, def]) => {
+								return [key, enableTypeCoercion(def, options)];
+							}),
+						),
+					}) as $ZodType<unknown, {}>;
+
+					object._zod.disc = def.options[index]?._zod.disc;
```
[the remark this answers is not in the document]
chimame: I was going to add a comment later, but forgot.
Thanks for pointing this out.

PR #909, opened 2025-04-18
feat: Set empty object when zodDiscriminatedUnion is specified.

taku-hatano:
Pull Request for #892.

Thank you for your checking.

PR #733, opened 2024-08-05
fix(conform-zod): enable zod object coercion

edmundhung:
Fix #391.

PR #911, opened 2025-04-19
fix(conform-valibot): support nested fields with only checkboxes

chimame:
## Overview

Valibot has also been modified to support validation of forms with nested checkboxes only, just like zod.
references: https://github.com/edmundhung/conform/pull/733

Zod has a `zodDiscriminatedUnion` schema, but valibot does not. Therefore, the same support as https://github.com/edmundhung/conform/pull/909 will not be performed.

### DIFF
--- a/.changeset/brave-timers-walk.md
+++ b/.changeset/brave-timers-walk.md
@@ -1,0 +1,5 @@
+---
+'@conform-to/valibot': patch
+---
+
+fix: support nested fields with only checkboxes
--- a/.changeset/brave-timers-walk.md
+++ b/.changeset/brave-timers-walk.md
@@ -1,5 +1,5 @@
 ---
-'@conform-to/valibot': patch
+'@conform-to/valibot': minor
 ---
 
-fix: support nested fields with only checkboxes
+feat: support nested fields with only checkboxes
--- a/packages/conform-valibot/coercion.ts
+++ b/packages/conform-valibot/coercion.ts
@@ -29,6 +29,13 @@ type EnableTypeCoercionOptions = {
 	) => CoercionFunction | null;
 };
 
+const OBJECT_SCHEMA_TYPES: string[] = [
+	'object',
+	'loose_object',
+	'strict_object',
+	'object_with_rest',
+];
+
 /**
  * Reconstruct the provided schema with additional preprocessing steps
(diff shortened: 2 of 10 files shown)


chimame:
@edmundhung Please do a code review. 🙏

.changeset/brave-timers-walk.md
```diff
@@ -0,0 +1,5 @@
+---
+'@conform-to/valibot': patch
+---
+
+fix: support nested fields with only checkboxes
```
edmundhung, line 5: Nitpick: I think this is more like a feature / minor. But up to you :)

edmundhung (approved):
Looks good to me. Now that we are planning to ship zod v4 support in a minor, should we fix the version of `@conform-to/valibot` so that it share the same version number as the rest of the packages? We can do that by updating the `fixed` options in the changeset config.

We will likely decouple the versions of each package in v2. Just thinking it might be better keep them consistent at this stage first. What do you think?

.changeset/brave-timers-walk.md
chimame: To be honest, I was worried about what to do. 😅
I just wanted to make a patch release based on past [pull requests](https://github.com/edmundhung/conform/pull/733).

chimame:
From this [comment](https://github.com/edmundhung/conform/pull/876#issuecomment-2713710194), I thought that `@conform-to/valibot` would be released as 1.0.0 and would be managed as a separate version. Therefore, I did not include `@conform-to/valibot` in the `fixed` option of `changesets`.

So my understanding was as follows. Sorry if I'm wrong.

- v1: Only `@conform-to/valibot` is managed as a separate version number, other packages are managed as the same version number
- v2: All packages are managed as separate version numbers

edmundhung:
> From this comment, I thought that @conform-to/valibot would be released as 1.0.0 and would be managed as a separate version.

Yes, that was the idea as I thought we will have to do a major bump for the zod package for v4 soon. But now that we have a better idea how to land zod 4 without a major bump, just wonder if we should keep the valibot package version consistent first and then communicate to people that the version number will be separated from v2 all together.

This is just an idea. Happy to keep it as is.

chimame:
I think it's okay to adopt your idea about version number management. Please let me know at any time if you want to change it. :+1:

PR #917, opened 2025-04-26
fix: where `undefined` value could not be correctly verified when `array` was defined in `nullish` or `optional` schema

### LINKED ISSUE #916: [conform-to-valibot] defaults for nullish fields are not respected

`parseWithValibot` does not work correctly with schemas using `nullish()` and defaults. 
The same schema and value being parsed with valibot directly comes back green. 

```ts
  describe('parseWithValibot()', () => {
    it.each([
      {}, 
      { ticketSpecs: [] }, 
      { ticketSpecs: null }
    ])(
      'falls back to default value if provided',
      (value) => {
        const schema = object({
          ticketSpecs: nullish(array(number()), []),
        });
        const fd = new FormData();
        Object.entries(value).forEach(([k, v]) => fd.append(k, v));
        const res = parseWithValibot(fd, { schema });
        expect(res.status).toBe('success'); // is error right now
      },
    );
  });
```

### Conform version

v1.0.1

I created a little test suite so you can easily reproduce it. 
It also provides proof that the same schema is working when using valibot directly.

https://github.com/edmundhung/conform/compare/main...mephju:conform:nullish-defaults-fail?expand=1

_No response_

### Screenshots or Videos

_No response_

### Additional context

_No response_

chimame:
## Overview

Close #916 

## Details

The following validation should succeed:

```ts
import { object, nullish, array, number } from 'valibot';
import { parseWithValibot } from '@conform-to/valibot';

const schema = object({
  age: nullish(array(number())),
});

expect(parseWithValibot(new FormData(), { schema })).toMatchObject({
  status: 'success',
  value: {},
});
```

However, the above validation does not pass because `@conform-to/valitob` uses `transform` to convert empty strings as follows:

```ts
import * as v from 'valibot';

const Schema = v.object({
  ticketSpecs: v.pipe(
    // The `unknown` schema makes it seem as though the `ticketSpecs` value must have some value.
    v.unknown(),
    v.transform(v => typeof v !== 'string' ? v : (v === '' ? undefined : v)),
    v.nullish(
      v.array(
        v.number()
      ),
    []),
  )
});

const result = v.safeParse(Schema, {});
console.log(result);
// => `result.success` is not true
```

As the comment in the code says, the `unkown` schema validates on the assumption that a value exists. The correct schema transformation should be as follows:

```ts
import * as v from 'valibot';

const Schema = v.object({
  ticketSpecs: v.nullish(
    v.pipe(
      v.unknown(),
      v.transform(v => typeof v !== 'string' ? v : (v === '' ? undefined : v)),
      v.nullish(
        v.array(
          v.number()
        ),
      []),
    )
  )
});

const result = v.safeParse(Schema, {});
console.log(result);
// => `result.success` is true
```

### DIFF
--- a/.changeset/kind-coats-lie.md
+++ b/.changeset/kind-coats-lie.md
@@ -1,0 +1,5 @@
+---
+'@conform-to/valibot': patch
+---
+
+fix: where `undefined` value could not be correctly verified when `array` was defined in `nullish` or `optional` schema
--- a/packages/conform-valibot/coercion.ts
+++ b/packages/conform-valibot/coercion.ts
@@ -215,13 +215,13 @@ function generateWrappedSchema<T extends GenericSchema | GenericSchemaAsync>(
 
 	// `expects` is required to generate error messages for `TupleSchema`, so it is passed to `UnkonwSchema` for coercion.
 	const unknown = { ...valibotUnknown(), expects: type.expects };
+	const default_ = 'default' in type ? type.default : undefined;
 	if (transformAction) {
 		const schema = type.async
 			? pipeAsync(unknown, transformAction, type)
 			: pipe(unknown, transformAction, type);
 
 		if (rewrap) {
-			const default_ = 'default' in type ? type.default : undefined;
 			return {
 				transformAction: undefined,
 				schema: type.reference(schema, default_),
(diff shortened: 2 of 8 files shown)


edmundhung (approved):

edmundhung:
> @edmundhung I apologize for asking for so many reviews. I found a problem and have fixed it. Please review the code. 🙏

No worries! Thanks for working on a fix quickly!

PR #918, opened 2025-04-27
docs: Add installation guide

chimame:
## Overview

Added documentation on installation for using Conform.

## Details

- To use Conform, you need to install the corresponding schema library and helper, so we have added detailed information.
- Since documentation for [using Zod v4](https://github.com/edmundhung/conform/pull/902#issuecomment-2816810238) will also be needed in the future, we are creating a new page.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -1,0 +1,102 @@
+# Installation
+
+To use Conform, you need to install it along with the schema library you want to support.
+
+## Zod
+
+If you are using [Zod](https://zod.dev/), install the validation helper library that supports Zod schemas.
+
+```bash
+npm install @conform-to/react @conform-to/zod zod
+```
+
+To validate FormData, use `parseWithZod`.
+
+```ts
+import { useForm } from '@conform-to/react';
+import { parseWithZod } from '@conform-to/zod';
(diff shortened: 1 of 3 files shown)


chimame:
@edmundhung 

I [commented](https://github.com/edmundhung/conform/pull/902#issuecomment-2816810238) on the documentation for installation and usage when Zod v4 is supported.
I have already created documentation on installation. I am thinking of adding it here when Zod v4 is supported. What do you think?

docs/installation.md
```diff
@@ -0,0 +1,102 @@
+# Installation
```
edmundhung, line 1: How about naming it `Schema Library` and put it on the Integration section next to the `UI Library` guide?

edmundhung (approved):
Just a minor comment, feel free to merge :)

docs/installation.md
chimame: Thank you for your advice. 👍 

Adding documentation to the section next to `UI Libraries` might be a good idea.
However, I think installation is the first thing users want to know about how to use it. So I added it to the section next to `Overview`.

I'll try adding it here first and then edit it if it's inappropriate.

PR #925, opened 2025-05-02
fix: re-support Valibot `config`/`fallback` methods for piped primitive value schemas

### LINKED ISSUE #924: [@conform-to/valibot] `config`/`fallback` methods don't work with piped primitive value schemas

When we call `parseWithValibot` function with a Valibot schema that contains `config`/`fallback` methods for piped primitive value schemas, the effect of `config`/`fallback` methods disappears.

### Conform version

@conform-to/valibot v1.0.1

Run the following test:

```javascript
import { describe, expect, test } from 'vitest';
import { config, empty, object, pipe, string } from 'valibot';
import { parseWithValibot } from '@conform-to/valibot';

describe('config', () => {
  test('should fail with message', () => {
    const schema = object({
      key: config(pipe(string(), empty()), { message: 'wrong value' }),
    });
    const input = new URLSearchParams({ key: 'value' });
    const output = parseWithValibot(input, { schema });
    expect(output).toMatchObject({
      status: 'error',
      error: { key: ['wrong value'] },
    });
  });
});
```

Actual output value: `{ status: 'error', error: { key: ['Invalid length: Expected 0 but received 5'] }, ... }`

Expected output value: `{ status: 'error', error: { key: ['wrong value'] }, ... }`

_No response_

### Screenshots or Videos

_No response_

### Additional context

This issue is a regression by chimame/conform-to-valibot#57 (since conform-to-valibot v1.14.2).

nanto:
Valibot `config`/`fallback` methods for piped primitive value schemas had worked until conform-to-valibot v1.14.1 but stop working since v1.14.2 (when chimame/conform-to-valibot#57 was merged). This pull request make them work again.

```javascript
// Example of config method for piped primitive value schema
const schema = v.object({
  key: v.config(
    v.pipe(v.number(), v.minValue(0), v.maxValue(10)),
    { message: 'Out of range' },
  ),
});
```

This pull request doesn't treat `config`/`fallback` methods for kinds of object schemas, which have not worked since before conform-to-valibot v1.14.1. (Maybe we have to modify Valibot internal data structure to support those cases.)

```javascript
// Example of fallback method for object schema
const schema = v.fallback(
  v.object({ key: v.number() }),
  { key: 0 },
);
```

Fixes #924


nanto:
This pull request reverts change of coerced schema structure (<https://github.com/chimame/conform-to-valibot/pull/57#issuecomment-2646435844>) since `config`/`fallback` information is lost by that change.

packages/conform-valibot/coercion.ts
```diff
@@ -285,19 +315,6 @@ function enableTypeCoercion<T extends GenericSchema | GenericSchemaAsync>(
 	transformAction: TransformAction<unknown, unknown> | undefined;
 	schema: GenericSchema | GenericSchemaAsync;
 } {
-	if ('pipe' in type) {
```
chimame, line 288: Here's the code that addresses the problem you pointed out in the original code. How about making the `if ('pipe' in type)` true case into a function? That way you don't need an `if` in the `switch`.

```diff
if ('pipe' in type) {
  const { transformAction, schema: coercedSchema } = enableTypeCoercion(
    type.pipe[0],
    options,
  );

+  if (transformAction) {
+    const unknown = { ...valibotUnknown(), expects: type.expects };
+    const schema = type.async
+      ? pipeAsync(unknown, transformAction, type)
+      : pipe(unknown, transformAction, type);
+
+    return { transformAction, schema };
+  }

  const schema = type.async
    ? pipeAsync(coercedSchema, ...type.pipe.slice(1))
    : // @ts-expect-error `coercedSchema` must be sync here but TypeScript can't infer that.
      pipe(coercedSchema, ...type.pipe.slice(1));

  return { transformAction, schema };
}
```

packages/conform-valibot/tests/coercion/schema/config.test.ts
```diff
@@ -0,0 +1,18 @@
+import { config, empty, object, pipe, string } from 'valibot';
```
chimame, line 1: Thanks for creating the test.
`fallback` and `config` are actions, not schemas, so could you put them in `tests/coercion/action/config.test.ts`?

chimame (commented):
Thank you for the pull request!
It's very helpful that you helped me create code to address the problem.

I also thought about the code to solve the problem, and I think it could be a little simpler. Can you check it?

packages/conform-valibot/coercion.ts
nanto: I see. Simple is better!

Creating schemas that may not be used can slow down execution time when pipes are nested many times ([I make a benchmark](https://gist.github.com/nanto/873606402bbdae027663a3d823a0bfae)). However, the case pipes are nested many times is not practical, I think.

packages/conform-valibot/tests/coercion/schema/config.test.ts
nanto: `fallback` and `config` are categorized as methods so I moved the test files to `tests/coercion/method/`.

By the way, since `test/coercion/method/` directory is created, it might be better to move some test cases in [`test/coercion/schema/wrap.test.ts`](https://github.com/edmundhung/conform/blob/33de60262d5cf73ed12761eed9d018d890d2e753/packages/conform-valibot/tests/coercion/schema/wrap.test.ts#L50-L86) and [`test/coercion/schema/wrapAsync.test.ts`](https://github.com/edmundhung/conform/blob/33de60262d5cf73ed12761eed9d018d890d2e753/packages/conform-valibot/tests/coercion/schema/wrapAsync.test.ts#L55-L108) into `test/coercion/method/pipe{,Async}.test.ts`.

nanto:
Thank you for your review! I modified codes and changed test file paths.

chimame (approved):
Thanks for the fix.
Let's merge this!

packages/conform-valibot/tests/coercion/schema/config.test.ts
chimame: You're right.
I'll move on to that in the future.

PR #941, opened 2025-05-24
chore: refine build setup

edmundhung:
- All packages will now output the built files to the `dist` folder
- Enabled `declarationMap`in tsconfig so that IDE can jump to the original source code during development.
- Removed `.npmignore` (blocklist) in favor of `files` on packages.json (allowlist)
- Make sure LICENSE is included in the published package


chimame (approved):
LGTM 👍
Thank you for aligning the release build. I'm very happy because it's something I was planning to do later.

With this change, it's easier to reference TypeScript files from test files, making unit testing easier.

PR #945, opened 2025-05-30
fix(conform-zod): `discriminatedUnion` schema to work properly in Zod versions 3.25.35 and later

### LINKED ISSUE #944: Error when using discriminatedUnion with parseWithZod v4

After upgrading to v4 I am experiencing problems when using the z.discriminatedUnion together with parseWithZod.

I'm on React Router v7.6.1 and parsing the formData within my action.

Error Message:

`TypeError: undefined is not an object (evaluating 'o._zod.propValues[def.discriminator]')
`

FormData:

`{
intent: "edit-item",
id: "23",
name: "Test"
}`

```
import { z } from 'zod/v4'
import { parseWithZod } from '@conform-to/zod/v4'

export const action = async ({ request, params }: Route.ActionArgs) => {
	
        const formData = await request.formData()

	const submission = parseWithZod(formData, {
		schema: z.discriminatedUnion('intent', [
			z.object({
				intent: z.literal('edit-item'),
				id: z.number(),
                                 name: z.string(),
			}),
			z.object({
				intent: z.literal('edit-order'),
				id: z.number(),
			}),
		]),
	})

       *----- ERROR -> parsing fails -------*

	if (submission.status !== 'success') {
		return data(
			{ result: submission.reply() },
			{ status: submission.status === 'error' ? 400 : 200 }
		)
	}

	console.log('Order edit submission:', submission.value)
}
```

### Conform version

1.6.0

1. Send formData to the RR action
2. Parse formData with a Zod schema using a discriminatedUnion
3. Log the submission result

Chrome

### Screenshots or Videos

Image

### Additional context

_No response_

chimame:
## Overview

Close #944 

## Details

Zod changed the internal property used by `discriminatedUnion` from version 3.25.35 onwards. As a result, schema validation using `discriminatedUnion` will not work properly in Zod versions 3.25.35 and later. Fixed so that it works properly in versions 3.25.35 and later.

https://github.com/colinhacks/zod/commit/fe75806cbb64683f237ed0e4ee93b67a2c5e976d

However, this is a workaround fix. I would like to use a process that does not actually need to access this internal property.

### DIFF
--- a/.changeset/grumpy-starfishes-invent.md
+++ b/.changeset/grumpy-starfishes-invent.md
@@ -1,0 +1,5 @@
+---
+'@conform-to/zod': patch
+---
+
+fix(conform-zod): `discriminatedUnion` schema to work properly in Zod versions 3.25.35 and later
--- a/packages/conform-zod/v4/coercion.ts
+++ b/packages/conform-zod/v4/coercion.ts
@@ -285,8 +285,10 @@ export function enableTypeCoercion<Schema extends $ZodType>(
 					}) as $ZodType<unknown, {}>;
 
 					// The discriminate key is obtained from the defined Object.
-					// If you regenerate the Object schema, the `disc` property disappears. Therefore, set the one obtained from the original Object.
-					// https://github.com/colinhacks/zod/blob/94871b708300ab67c4964025354c5199d335e55a/packages/zod/src/v4/core/schemas.ts#L2055-L2070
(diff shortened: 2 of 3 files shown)


chimame:
@edmundhung 

I looked at Zod's code and couldn't find any other good solution. If you have any suggestions, please let me know. 🙏

PR #952, opened 2025-06-05
Expose value in Submission object during successful field validation

lhapaipai:
First of all, thank you for your amazing work on Conform.

## Context

Currently, `parseWithZod` returns a Submission object whose value is only available when `status === 'success'`, i.e., after a full form submission.

However, during field-level validation (e.g., with `shouldValidate: 'onInput'`), even when the data is valid, status is undefined, and the transformed value is not exposed.
Problem

This behavior prevents access to Zod-transformed values during successful partial validation.

As a result, it's difficult to use those values in real-time features such as a live preview component.

This PR follows up on the discussion: https://github.com/edmundhung/conform/discussions/898

## Proposal

Update the createSubmission function to include the value property in the returned object when `context.value` is defined, even if status is undefined.

## Benefits

- Improved reactivity: Enables real-time updates with already-validated 
- Avoids redundant parsing: No need to manually re-parse with safeParse inside onValidate callbacks.
- Maintains backward compatibility: The semantics of status remain unchanged.

## Example

I’ve created a playground at /preview to demonstrate this behavior and test how the value behaves during partial field validation.

## Conclusion

This change enhances the developer experience by providing consistent access to validated data without disrupting existing API behavior.

### DIFF
--- a/packages/conform-valibot/tests/coercion/schema/looseObject.test.ts
+++ b/packages/conform-valibot/tests/coercion/schema/looseObject.test.ts
@@ -17,7 +17,6 @@ describe('looseObject', () => {
 		input1.append('key3', '');
 		const output2 = parseWithValibot(input1, { schema: schema1 });
 		expect(output2.status).toBe('success');
-		// @ts-expect-error
 		expect(output2.value).toStrictEqual({
 			key1: 'test',
 			key2: 123,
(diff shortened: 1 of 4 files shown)


edmundhung:
Hi @lhapaipai,

Thanks so much for the PR, and sorry for the late response. I'm slightly hesitant to make this change to the current `parse` helper, as we are close to releasing our v2 APIs—where the existing `parse` helper will be replaced with a much simpler version, giving you full control over parsing and validation. For example:

```ts
const submission = parse(formData); // <- No schema passed
const result = schema.safeParse(submission.payload); // <- Validate the submission yourself
```

That said, your suggestion is safe and doesn’t cause any issues, so I'm open to feedback here. Let me know if you have any thoughts.

edmundhung:
Hey, thanks so much for taking the time to put this PR together. Now that all future v2 APIs are already available, I think the better path forward is to migrate over to the new APIs which provide more explicit control over parsing and value access. Since we are trying to avoid making changes to v1 at this point, I'm going to close this PR. 

If you are still running into similar issues with the v2 APIs, please don't hesitate to open a new discussion or issue!

PR #967, opened 2025-06-19
chore: setup both browser and node tests with vitest

edmundhung:
This PR updated Vitest version to v3 with setup for both browser and node environments:

- If the test file matches `*.node.test.{ts,tsx}`, it will run on node environment
- If the test file matches `*.browser.test.{ts,tsx}`, it will run on browser environment
- Otherwise, the test file will run on both environments (i.e. `*.test.{ts,tsx}`)

I implemented this initially through a inline projects config in individual packages folder but vitest doesn't support nested projects as I tested it. 

cc: @chimame


edmundhung:
Good call. How can I missed those tests 😅 I have renamed all conform-zod tests in https://github.com/edmundhung/conform/pull/967/commits/bcb678c956438fdf7752e21ceeaefb71b52a1714. We still have some legacy tests left but we can remove those later.

PR #983, opened 2025-07-11
refactor: tighten path validation when parsing formdata

edmundhung:
To prepare for upcoming changes, we have overhauled the way paths are parsed and formatted:

- The parser now detects invalid syntax sooner and throws an error immediately on any malformed path.
- When parsing formdata, parse errors are caught and suppressed—any entry whose path is invalid will simply be omitted from the final submission.

This shouldn't have any impact to existing users. Users who rely on inferred names from field metadata, or who supply explicit names using dot-and-bracket notation, will see no change in behavior.


packages/conform-valibot/tests/coercion/schema/array.test.ts
```diff
@@ -24,14 +24,14 @@ describe('array', () => {
...
-		const formData2 = createFormData('nest[].name', 'test name');
+		const formData2 = createFormData('nest[0].name', 'test name');
 		const output2 = parseWithValibot(formData2, { schema: schema2 });
 		expect(output2).toMatchObject({
 			status: 'success',
 			value: { nest: [{ name: 'test name' }] },
 		});
 
-		const errorFormData = createFormData('nest[].name', '');
+		const errorFormData = createFormData('nest[0].name', '');
```
edmundhung, line 34: @chimame FYI: you may need to update the tests in your repo. We originally added empty-bracket support to make it easier to resolve validation attributes from the derived constraints, but empty brackets aren't officially supported when parsing formdata as it is ambiguous which item you are referring to in a nested object.

With this refactor, we have added limited empty-bracket support: if we see `items[]` at the **end** of a path, we will initialize an array and append the value. But empty brackets anywhere else in the path remain unsupported.

packages/conform-dom/index.ts
```diff
@@ -34,10 +31,7 @@ export {
 } from './submission';
 export {
 	getFormData,
-	getPaths,
-	formatPaths,
+	getPathSegments as getPaths,
+	formatPathSegments as formatPaths,
```
edmundhung, line 35: I have renamed several helpers internally. But still exported in the old names with the exact same signature. So it shouldn't breaking any people.

packages/conform-dom/index.ts
```diff
@@ -16,12 +16,9 @@ export {
 export {
 	type FieldElement,
 	isFieldElement,
+	isGlobalInstance,
 	updateField as unstable_updateField,
 	createFileList,
-	createGlobalFormsObserver as unstable_createGlobalFormsObserver,
-	focus as unstable_focus,
-	change as unstable_change,
-	blur as unstable_blur,
```
edmundhung, line 24: I moved all these exports to the future export so it's easier to manage as I put up more helpers that are only used by those future APIs.

packages/conform-dom/tests/formdata.test.ts
```diff
@@ -0,0 +1,756 @@
...
+	expect(() => getPathSegments('[a.b]')).toThrow(
+		'Invalid path syntax at position 0 in "[a.b]"',
+	);
+	expect(() => getPathSegments('a[b].b')).toThrow(
+		'Invalid path syntax at position 1 in "a[b].b"',
+	);
+	expect(() => getPathSegments('a[[]]')).toThrow(
+		'Invalid path syntax at position 1 in "a[[]]"',
+	);
+});
```
edmundhung, line 64: This should make it clear what kind of paths we accept.

packages/conform-dom/tests/formdata.test.ts
```diff
@@ -0,0 +1,756 @@
...
+		expect(() => setValueAtPath(target, 'foo..bar', 'test')).toThrowError();
+		expect(() =>
+			setValueAtPath(target, 'foo..bar', 'test', { silent: true }),
+		).not.toThrowError();
+		expect(() =>
+			setValueAtPath(target, 'foo.bar[].baz', 'test'),
+		).toThrowError();
+		expect(() =>
+			setValueAtPath(target, 'foo.bar[].baz', 'test', { silent: true }),
+		).not.toThrowError();
```
edmundhung, line 338: Re: empty bracket support

packages/conform-valibot/tests/coercion/schema/array.test.ts
chimame: Thank you for letting me know. 👍 
I'll keep that in mind when I create tests in the future. I'll also try to change any existing tests that seem to need changing.

Issue #997, opened 2025-08-08
【conform-to/zod】getZodConstraint returns different constraints for z.transform between v3 and v4


taku-hatano:
### Describe the bug and the expected behavior

Hello team,

In `getZodConstraint `, v3 and v4 returns different constraints.
I think it should return same constraints.

See `Steps to Reproduce the Bug or Issue`.

### Conform version

v1.8.2

### Steps to Reproduce the Bug or Issue

1. generate zod schema like below:
```ts
const schema = z
	.object({
		transform: z
			.object({
				key: z.string(),
			})
			.transform((val) => val),
	})
```

2. getZodConstraint by v3
```ts
import { getZodConstraint } from "@conform-to/zod/v3";
const constraint = getZodConstraint(schema);
console.log(constraint)
```

This return:
```ts
{
	transform: {
		required: true,
	},
	'transform.key': {
		required: true,
	},
}
```

3. getZodConstraint by v4
```ts
import { getZodConstraint } from "@conform-to/zod/v4";
const constraint = getZodConstraint(schema);
console.log(constraint)
```

This return:
```ts
{
	transform: {
		required: true,
	},
}
```

PR #998, opened 2025-08-08
fix(conform-zod/v4): getZodConstraint should resolve ZodPipe similar to v3

taku-hatano:
Pull Request for #997 

I also added tests for methods that return ZodPipe other than `transform`.
Thank you for your checking.


edmundhung (approved):
Thank you! Looks good to me from the behavior perspective. But I will defer to @chimame for reviewing the implementation :)

PR #1006, opened 2025-08-27
fix(conform-dom): Metadata cannot be retrieved correctly when an array of objects nested two or more levels deep is defined

### LINKED ISSUE #1005: `required` becomes undefined for fields with more than 2 levels of depth

Here is the problem:
- `fields.subFields.required` → `boolean`
- `fields.subFields.subSubFields.required` → `undefined` (❌ should be `boolean`)

Environment:
- React Router 7
- Zod 4
- Conform 1.8.2

### Conform version

v1.8.2

Create a Zod 4 Schema with 2 levels of depth:

```tsx
const Schema = z.object({
  isRequired: z.string(),
  items: z.array(
    z.object({
      isRequired: z.string(),
      subItems: z.array(
        z.object({
          shouldBeRequired: z.string(),
        }),
      ),
    }),
  ),
});
```

Then, you will see:

```tsx
  id: "my-form",
  constraint: getZodConstraint(Schema),
  onValidate({ formData }) {
    return parseWithZod(formData, {
      schema: Schema,
      formatError: (issues: z.core.$ZodIssue[]) => formatError(t, issues),
    });
  },
  defaultValues: {
    isRequired: "",
    items: [
      {
        isRequired: "",
        subItems: [
          {
            shouldBeRequired: "",
          },
        ],
      },
    ],
  },
  shouldValidate: 'onBlur',
  shouldRevalidate: 'onInput',
});

console.log({
  isRequired: fields.isRequired.required, // true
  items: {
    isRequired: fields.items.getFieldList()[0]?.getFieldset().isRequired.required, // true
    subItems: {
      shouldBeRequired: fields.items.getFieldList()[0]?.getFieldset().subItems.getFieldList()[0]?.getFieldset().shouldBeRequired, // undefined
    },
  }
});
```

Chrome

### Screenshots or Videos

_No response_

### Additional context

_No response_

chimame:
## Overview

Fix #1005

### DIFF
--- a/packages/conform-dom/form.ts
+++ b/packages/conform-dom/form.ts
@@ -519,7 +519,7 @@ function createConstraintProxy(
 					// This overrides the current number segment with an empty string
 					// which will be treated as an empty bracket
 					path[i] = '';
-					break;
+					continue;
 				}
 			}
 
--- a/.changeset/forty-flies-love.md
+++ b/.changeset/forty-flies-love.md
@@ -1,0 +1,5 @@
+---
+'@conform-to/dom': patch
+---
+
+fix(conform-dom): Metadata cannot be retrieved correctly when an array of objects nested two or more levels deep is defined


edmundhung (approved):

packages/conform-dom/form.ts
```diff
@@ -519,7 +519,7 @@ function createConstraintProxy(
 					// This overrides the current number segment with an empty string
 					// which will be treated as an empty bracket
 					path[i] = '';
-					break;
+					continue;
```
edmundhung, line 522: I just had another look and I am not sure why this change fixes the issue actually.

The original `break` was intentional. When resolving a constraint we progressively generalise the field name and scan each alternative in order:

```
array[0].anotherArray[1].key
→ array[0].anotherArray[].key
→ array[].anotherArray[].key
```

Switching to `continue` skips the intermediate variant and jumps straight from the fully specific name to the fully general one here. 🤔

edmundhung: Hmmm, just did a quick test with the `createStateProxy` and it seems to be working as expected:

```js
function createStateProxy(fn) {
	const cache = {};
	return new Proxy(cache, {
		get(_, name, receiver) {
			if (typeof name !== 'string') {
				return;
			}

			return (cache[name] ??= fn(name, receiver));
		},
	});
}

let t = createStateProxy((name, proxy) => {
  console.log('Looking up', name);
  if (name) {
    return proxy[name.slice(0, -1)];
  }

  console.log('done');
});

t.abc

// > Looking up abc
// > Looking up ab
// > Looking up a
// > Looking up 
// > done
```

chimame: Thanks for looking into it.
I didn't realize it was fixed in #983 . Let's close this PR then.

PR #1039, opened 2025-09-15
fix(conform-react/future): use static reset key for initial form state

### LINKED ISSUE #1030: generateUniqueKey in react/future causes errors in Next.js unless wrapped in Suspense

Using `@conform-to/react/future` in a Next.js with Dynamic IO enabled (now called "Cache Components") causes the following error in dev mode and results in failed production builds:

> [!CAUTION]
> Error: Route "*{route path}*" used `Date.now()` inside a Client Component without a Suspense boundary above it. See more info here: https://nextjs.org/docs/messages/next-prerender-current-time-client

I've tracked the cause of the error to `generateUniqueKey()` in `utils.ts`.

It uses two APIs that are considered dynamic under DIO: `Date.now` & `Math.random`.

To fix the problem the function should only be called on the client somehow or replaced by something else.

I found two places where it is used:
1. as the `resetKey` in `initializeState` (file: `state.ts`)
2. in the `validate.onUpdate` function of the `actionHandlers` (file:`intent.ts`)

One suggestion is to try replacing *generateUniqueKey* with `useId` for the *resetKey*. Since this looks like it's called immediately on the first render (on the SSR pass as well), I suspect that will fix the problem.

## Workaround

Wrap the Form in a `Suspense` boundary. However this causes the form to be hidden on initial page load and makes it no longer possible to work with JS disabled which is very unfortunate.

### Conform version

v1.9.1

1. Start a Next.js project

```bash
pnpm create next-app@canary my-app --yes
cd my-app
pnpm dev
```

2. Enable "Cache Components" by modifying next.config.mjs

```js
config.experimental = {};
config.experimental.cacheComponents = true;
```

3. Add a form component anywhere and inspect the console (might need a full refresh).

Chrome

### Screenshots or Videos

Image

### Additional context

_No response_

edmundhung:
Fix #1030
Close #1035

This is originally submitted in #1035 by @apostolos, but I can't push any changes to it as it is made from the default branch of a fork.

### DIFF
--- a/.changeset/flat-balloons-tap.md
+++ b/.changeset/flat-balloons-tap.md
@@ -1,0 +1,5 @@
+---
+'@conform-to/react': patch
+---
+
+Use static reset key for initial form state to fix Next.js prerendering hydration issues
--- a/packages/conform-react/future/hooks.tsx
+++ b/packages/conform-react/future/hooks.tsx
@@ -133,7 +133,7 @@ export function useConform<ErrorShape, Value = undefined>(
 ): [FormState<ErrorShape>, (event: React.FormEvent<HTMLFormElement>) => void] {
 	const { lastResult } = options;
 	const [state, setState] = useState<FormState<ErrorShape>>(() => {
-		let state = initializeState<ErrorShape>();
+		let state = initializeState<ErrorShape>('initial');
 
 		if (lastResult) {
 			state = updateState(state, {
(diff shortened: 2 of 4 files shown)


edmundhung:
This is landed in [v1.10.1](https://github.com/edmundhung/conform/releases/tag/v1.10.1).

PR #1077, opened 2025-10-25
chore: update vitest and vitest/browser dependencies to latest versions

chimame:
## Overview

https://vitest.dev/guide/migration.html#vitest-4


edmundhung (approved):
Thank you!

PR #1079, opened 2025-10-25
feat(conform-react/future): support resetting form to specific value

edmundhung:
Inspired by https://github.com/edmundhung/conform/discussions/1014#discussioncomment-14774863

This allows resetting the form with a specific value from the server:

```tsx
// Reset form to a specific value after submission
return {
  result: report(submission, {
    reset: true,
    value: updatedData,
  }),
};
```

This is also useful for React Router app where you want to reset the form to the last saved value after a successful submission, without having to use additional tricks to delay when form reset happens in the client:

### DIFF
--- a/.changeset/silly-moose-explain.md
+++ b/.changeset/silly-moose-explain.md
@@ -1,0 +1,13 @@
+---
+'@conform-to/react': minor
+---
+
+Form metadata now exposes `defaultValue`. As default values can be set through the `useForm` hook, reset intents, or server result, this provides clarity on the current baseline for features like dirty checking.
+
+```tsx
+const form = useFormMetadata();
+const dirty = useFormData(
+  form.id,
+  (formData) => isDirty(formData, { defaultValue: form.defaultValue }) ?? false,
+);
+```
--- a/docs/api/react/future/useFormMetadata.md
+++ b/docs/api/react/future/useFormMetadata.md
@@ -58,6 +58,10 @@ Form-level validation errors, if any exist.
 
 Object containing errors for all touched fields.
 
+### `defaultValue: Record<string, unknown>`
+
+Initial form values.
+
 ### `props: FormProps`
 
 Form props object for spreading onto the `<form>` element:
--- a/examples/react-router/app/routes/todos.tsx
+++ b/examples/react-router/app/routes/todos.tsx
@@ -68,7 +68,7 @@ export default function Example({
 	const dirty = useFormData(
 		form.id,
 		(formData) =>
-			isDirty(formData, { defaultValue: loaderData.todos }) ?? false,
+			isDirty(formData, { defaultValue: form.defaultValue }) ?? false,
 	);
 	const tasks = fields.tasks.getFieldList();
 
--- a/packages/conform-react/future/hooks.tsx
+++ b/packages/conform-react/future/hooks.tsx
@@ -174,7 +174,10 @@ export function useConform<
 ): [FormState<ErrorShape>, (event: React.FormEvent<HTMLFormElement>) => void] {
 	const { lastResult } = options;
 	const [state, setState] = useState<FormState<ErrorShape>>(() => {
-		let state = initializeState<ErrorShape>({ resetKey: INITIAL_KEY });
+		let state = initializeState<ErrorShape>({
+			defaultValue: options.defaultValue,
+			resetKey: INITIAL_KEY,
+		});
 
 		if (lastResult) {
 			state = updateState(state, {
@@ -187,8 +190,8 @@ export function useConform<
 					handlers: actionHandlers,
 					reset: (defaultValue) =>
 						initializeState<ErrorShape>({
+							defaultValue: defaultValue ?? options.defaultValue,
 							resetKey: INITIAL_KEY,
-							defaultValue,
 						}),
 				},
 			});
@@ -222,7 +225,9 @@ export function useConform<
 					ctx: {
 						handlers: actionHandlers,
 						reset(defaultValue) {
-							return initializeState<ErrorShape>({ defaultValue });
+							return initializeState<ErrorShape>({
+								defaultValue: defaultValue ?? optionsRef.current.defaultValue,
+							});
 						},
 					},
 				}),
@@ -263,9 +268,13 @@ export function useConform<
 		// Reset the form state if the form key changes
 		if (options.key !== keyRef.current) {
 			keyRef.current = options.key;
-			setState(initializeState<ErrorShape>());
+			setState(
+				initializeState<ErrorShape>({
+					defaultValue: optionsRef.current.defaultValue,
+				}),
+			);
 		}
-	}, [options.key]);
+	}, [options.key, optionsRef]);
 
 	useEffect(() => {
 		const formElement = getFormElement(formRef);
@@ -275,12 +284,12 @@ export function useConform<
 			resetKeyRef.current = state.resetKey;
 			resetFormValue(
 				formElement,
-				state.serverValue ?? optionsRef.current.defaultValue ?? {},
+				state.defaultValue,
 				optionsRef.current.serialize,
 			);
 			pendingValueRef.current = undefined;
 		}
-	}, [formRef, state.resetKey, state.serverValue, optionsRef]);
+	}, [formRef, state.resetKey, state.defaultValue, optionsRef]);
 
 	useEffect(() => {
 		if (state.targetValue) {
@@ -493,7 +502,7 @@ export function useForm<
 	fields: Fieldset<FormShape, ErrorShape>;
 	intent: IntentDispatcher<FormShape>;
 } {
-	const { id, defaultValue, constraint } = options;
+	const { id, constraint } = options;
 	const globalOptions = useContext(GlobalFormOptionsContext);
 	const optionsRef = useLatest(options);
 	const globalOptionsRef = useLatest(globalOptions);
@@ -573,7 +582,6 @@ export function useForm<
 		() => ({
 			formId,
 			state,
-			defaultValue: defaultValue ?? null,
 			constraint: constraint ?? null,
 			handleSubmit: handleSubmit,
 			handleInput(event) {
@@ -648,7 +656,6 @@ export function useForm<
 		[
 			formId,
 			state,
-			defaultValue,
 			constraint,
 			handleSubmit,
 			intent,
--- a/packages/conform-react/future/state.ts
+++ b/packages/conform-react/future/state.ts
@@ -31,8 +31,9 @@ export function initializeState<ErrorShape>(options?: {
 	return {
 		resetKey: options?.resetKey ?? generateUniqueKey(),
 		listKeys: {},
+		defaultValue: options?.defaultValue ?? {},
 		targetValue: null,
-		serverValue: options?.defaultValue ?? null,
+		serverValue: null,
 		serverError: null,
 		clientError: null,
 		touchedFields: [],
@@ -164,7 +165,7 @@ export function getDefaultValue(
 	const value = getValueAtPath(
 		context.state.serverValue ??
 			context.state.targetValue ??
-			context.defaultValue,
+			context.state.defaultValue,
 		name,
 	);
 	const serializedValue = serialize(value);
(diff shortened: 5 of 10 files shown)


edmundhung:
> No overload matches this call.
>   The last overload gave the following error.
>     Type '{ hello?: string | undefined; }' is not assignable to type 'Record<string, FormValue>'.
>       Property 'hello' is incompatible with index signature.
>         Type 'string | undefined' is not assignable to type 'FormValue'.
>           Type 'undefined' is not assignable to type 'FormValue'.ts(2769)

This is another `exactOptionalPropertyTypes` type issue 😅 I will get it fixed next

> The needed type feels like it should be structurally identical to DefaultValue.

This is a challenging topic. As I have no idea how the data is serialized when they are sent back to the client, my current type restricts it so that it accepts types that are JSON serializable only. It might not be an actual issue with React Router and Nextjs as they both uses a custom serializer these days. 

> if the form has an optional file input,

This is not actually a regression but just an edge case get uncovered once I fix the bugs with  `intendedValue` 😅

Update form value in Conform is a two step process:
1. Update defaultValue of each fields so your component can render new fields with related to the updates
2. Update the field value of each input elements in the DOM so that the value of existing fields get updated too

Before the bug is fixed, Conform did (1) only when you specify `intendedValue` without `reset`. So it works for you if you are using it to render new fields / unmount existing fields. But now that (2) happens, it will try to update the file input value, but it has no idea what the File should be as files are stripped from the server result. I was planning to leave this as an edge case that I won't cover in this PR, but maybe I need to fix it now...

edmundhung:
Awesome. Then I think this feature is good to go. Is the new name okay? I am also thinking if `updateValue` or `targetValue` is a better name..

transparent-citizen:
Inputs in React use `defaultValue` and `value` to reflect controlledness. I'm not convinced that this is the notion to convey here, so maybe it's better to avoid the naming similarity and better distinguish conform-isms from React-isms. Something more unique like `updateValue`, `targetValue`, `newValue`, `nextValue` might work.

`nextValue` still makes the most sense to me, but I'm just as happy working with something else if you have a preference! :+1:

transparent-citizen:
I've been testing the commits in this PR a little more thoroughly. Hope you find this overview useful:

<details>
<summary>Test case</summary>

```ts
import { parseSubmission, report } from "@conform-to/dom/future";
import { useForm } from "@conform-to/react/future";
import { coerceFormValue } from "@conform-to/zod/v4/future";
import { Form } from "react-router";
import z from "zod";

import type { Route } from "./+types/conform";

	z.object({
		fruit: z.string().optional(),
	}),
);

export const action = async ({ request }: Route.ActionArgs) => {
	const submission = parseSubmission(formData);

	const result = schema.safeParse(submission.payload);

	if (!result.success) {
		return {
			lastResult: report(submission, {
				error: {
					issues: result.error.issues,
				},
			}),
		};
	}

	return {
		lastResult: report(submission, {
			value: {
				fruit: "cherry",
			},
			reset: true,
		}),
	};
};

export default ({ actionData }: Route.ComponentProps) => {
	const { form, fields } = useForm({
		...actionData,
		defaultValue: {
			fruit: "apple",
		},
		schema,
	});

	return (
		<Form {...form.props} method="POST">
			<select
				defaultValue={fields.fruit.defaultValue}
				name={fields.fruit.name}
			>
				<option value="">Select a fruit</option>
				<option value="apple">apple</option>
				<option value="banana">banana</option>
				<option value="cherry">cherry</option>
			</select>
			<button type="submit">Submit</button>
		</Form>
	);
};
```
</details>

Commit 1, [`ef6e4a2`](https://github.com/edmundhung/conform/runs/53669052439) (`intendedValue`)
| reset | result |
| - | - |
| `true` | ❌ form updates to `defaultValue` |
| `false` | ✅ form updates to `intendedValue` |

Commit 2, [`d58bbce`](https://github.com/edmundhung/conform/pull/1079/commits/d58bbce9aa964d9646424b6bf0220763e67545ce) (`intendedValue`)
| reset | result |
| - | - |
| `true` | ❌ form updates to `defaultValue` |
| `false` | ✅ form updates to `intendedValue` |

Commit 3, https://github.com/edmundhung/conform/pull/1079/commits/d4e30c40a60c2d2d4c3be976ad4fcf4f3a6e345b (`value`)
| reset | result |
| - | - |
| `true` | ❌ form updates to `defaultValue` |
| `false` | ✅ form updates to `value` |

edmundhung:
@transparent-citizen Thanks a lot for giving it a thorough check! I can reproduce the issue with select resetting to the `defaultValue` when reset is `true`. But I couldn't reproduce your issue with the optional file schema. What error do you see?

edmundhung:
> To me it looks like <input> is the oddball, since updating defaultValue will rerender an input with a new value.

Yeah, I have been weaving between whether how much Conform should rely on React. There are things React could helps but the whole story with uncontrolled inputs is such a mess. 

> A hacky way for conform to update the fields when defaultValue updates would be to make key

Conform v1 (before v1.3) uses `key` exactly for this purpose. But there are a few issues:
- React 19 now warns if you try to spread props with key on an element. 
- If the key is changed while users focus is on the input, it triggers an on blur event which mess with the validation setup

transparent-citizen:
Hope your vacation allowed for some recovery!

> Conform v1 (before v1.3) uses `key` exactly for this purpose. But there are a few issues:

Thanks for pointing that out. Even if `key` had none of these issues, using it would still prevent the user from providing their own `key` to form elements, which might hinder optimization.

I wonder if the pain can be alleviated by remodeling `BaseMetadata` to provide a `value` and `onChange`, not just `defaultValue`. React users are already aware that they must choose between

- `value` + `onChange` (controlled), or
- `defaultValue` (uncontrolled)

Could this distinction not save conform from manually updating the DOM, as it offloads form value updates to React?

edmundhung:
> Hope your vacation allowed for some recovery!

edmundhung:
I am not sure if this case worths fixing though. Do we consider the previous submission cancelled? The action might have completed and updated your database with the last submission. Should we just recommend developer to disable the fields instead when a submission happens in this case? What if the submission is successful and we are just trying to reset the form with a new default value?

transparent-citizen:
> Do you mean the input and select shows a different value? e.g. `<input>` being 100 but `<select>` is showing 97.

Yes, that's exactly what's happening on my end.

This sequence diagram displays how I think the communication goes. Blue squares mean the user is typing.

image

Note that the end result ("hello") is what the user meant to type, but the user sees:

1. he
2. hello
3. he
4. hello

The way I understand it, react-router making use of `AbortController` has the ability to abort *ongoing* fetch requests. There are cases where the request has already finished, but the `lastResult` it contains is guaranteed to be stale:

image

This case occurs when on the client, the user has progressed the form state, and any `lastResult` pending processing would be destructive to the input, as it would remove characters. In the sequence diagram above, "hello" is set back to "he" the way you already described it, later mangled to "heworld".

As shown above, there exists a sequence of events in which state can become stale without `AbortController` having a chance to catch that.

> Are you suggesting that Conform should ignore the result if the value has changed since the last submission?

Yeah, exactly. The timestamps/logical clocks/sequence numbers in the pseudocode above were meant to say that conform might benefit from telling the apart stale from fresh state by making state sortable, as exchanging state under latency ascribes a lot of significance to its recency.

I was using react-router 7.9.5 in my tests above. I will later try and set up a test case that's easier to reproduce.

transparent-citizen:
Just so that you're aware, [d9e2227](https://github.com/edmundhung/conform/pull/1079/commits/d9e222726ee20fab616f319cafaec4e727416ab8) uncovers another `exactOptionalPropertyTypes` type issue, as it now complains when passing an undefined `onSubmit` to `useForm`.

edmundhung:
> As shown above, there exists a sequence of events in which state can become stale without AbortController having a chance to catch that.

I understand what you are refer to in theory. But it seems to me this is really unlikely to happen as the JS latency should be very low. I have a draft solution that skips the submission result if the submission payload doesn't match the current form data, but I am a bit hesitant it's worth the extra effort validating that. 🤔 

> this seems to cause an input event, meaning that it submits again, causing an infinite request loop

That's actually a new feature in v2 which I will need to document properly: Everytime Conform updates an input value and the value is actually changed, it will dispatch both change and input event to give you a chance to subscribe to it. To fix the issue you see, all you need to do is to skip form submission if the event default is prevented:

```diff
const submitForm: FormEventHandler<HTMLInputElement> = (event) => {
-  if (!event.currentTarget.form) {
+  if (!event.currentTarget.form || event.defaultPrevented) {
    return;
  }

  submit(event.currentTarget.form, {
    replace: true,
  });
};
```

transparent-citizen:
> That's actually a new feature in v2 which I will need to document properly

Ah, thank you very much! 🎉 This solves the request loop.

> But it seems to me this is really unlikely to happen as the JS latency should be very low.

JS latency can be significant if a render isn't cheap. I managed to make this effect more noticeable in a better test case.

The following setup features an `<input>` with `onInput={submitForm}` again, but in addition to that, it also renders 4,000 random numbers whenever the input changes. This is to simulate a search page with a long result list.

You can try and enter any sentence at a fast typing speed. In my tests, conform would reliably mess up the input when trying to apply `targetValue`.

**Test case**
<details>
<summary>Code</summary>

```ts

import { parseSubmission, report } from "@conform-to/dom/future";
import { useForm } from "@conform-to/react/future";
import { coerceFormValue } from "@conform-to/zod/v4/future";
import type { FormEventHandler } from "react";
import { Form, useSubmit } from "react-router";
import z from "zod";

import type { Route } from "./+types/conform";

	z.object({
		input: z.string().optional(),
	}),
);

export const loader = async ({ request }: Route.ActionArgs) => {
	const searchParams = new URL(request.url).searchParams;
	const submission = parseSubmission(searchParams);
	const result = schema.safeParse(submission.payload);

	if (!result.success) {
		throw result.error;
	}

	return {
		lastResult: result.data.input
			? report(submission, {
					targetValue: {
						input: result.data.input,
					},
				})
			: null,
	};
};

export default ({ loaderData }: Route.ComponentProps) => {
	const { form, fields } = useForm({
		...loaderData,
		schema,
	});

	const submit = useSubmit();

	const submitForm: FormEventHandler<HTMLInputElement> = (event) => {
		if (!event.currentTarget.form || event.defaultPrevented) {
			return;
		}

		submit(event.currentTarget.form, {
			replace: true,
		});
	};

	return (
		<Form {...form.props}>
			<input
				name={fields.input.name}
				onInput={submitForm}
			/>
			<button type="submit">Submit</button>
			{new Array(4000).fill(0).map((_, i) => (
				<div key={i}>{Math.random()}</div>
			))}
		</Form>
	);
};

```
</details>

transparent-citizen:
Here's a [StackBlitz](https://stackblitz.com/edit/github-xql5392t?file=app%2Froutes%2Fhome.tsx) to reproduce above test case.

If you reproduce it locally, the following manual intervention is not necessary. But on StackBlitz, hit submit once before trusting the application to be properly initialized.

Once initialized, enter any text to observe the text mangling.

My best guess of the root cause is that a stale `targetValue` is used to compute an upcoming form state.

> skips the submission result if the submission payload doesn't match the current form data

I wonder if this check really establishes freshness. One dimension getting lost in this comparison is the aspect of time. If you have two states and need to decide which one is fresher, it would require a property that allows to tell order. If every form state (current form data, payload or updated through intent) had an order property, it could be used to compare the current form data to an incoming payload and see if the current form data is already in a newer state.

edmundhung:
Thanks @transparent-citizen. your feedback has been really helpful and I appreciate it. Let's open a separate issue for this. 👍🏼

andrew-shpak:
No erros, so by default code be with `lastResult: undefined` whatever initial form state I have ?
This is clarification on my side

edmundhung:
@andrew-shpak So is that sorted now? or do you still think we shouldn't require either one. Just to make sure I don't miss your usecases :)

transparent-citizen:
I have a few similar cases. `ExampleForm` accepts a `lastResult` if it renders after an action. The parameter is optional, because it might also render before an action:

```ts
export const ExampleForm = (props: {
	lastResult?: SubmissionResult;
	// ...	
}) => {
	const { fields, form } = useForm(
		schema,
		{
			lastResult: null, // confusing, but necessary
			...props,
		},
	);
	// ...
};
```

Prior to 1.14.0, this would have been enough:

```ts
export const ExampleForm = (props: {
	lastResult?: SubmissionResult;
	// ...	
}) => {
	const { fields, form } = useForm({
		...props,
		schema
	});
	// ...
};
```

edmundhung:
See https://github.com/edmundhung/conform/pull/1116

PR #1090, opened 2025-11-09
fix(conform-zod): import from zod/v3 in conform-zod/v3

kesoji:
# Versions
- zod 4.1.12
- conform 1.13.3

# Problem 
importing from zod/v3 and conform-to/zod/v3 throws type error.

# Detail

This
```typescript
import { Welcome } from "../welcome/welcome";
import { useForm } from '@conform-to/react';
import { parseWithZod } from '@conform-to/zod/v3'; // <--
import { z } from 'zod/v3'; // <--

const schema = z.object({
  email: z.string(),
  password: z.string(),
});

export default function Home() {
  useForm({
    onValidate({ formData }) {
      return parseWithZod(formData, { schema });
    },
  });
  return <Welcome />;
}
```

throws
```
app/routes/home.tsx:14:39 - error TS2769: No overload matches this call.
  The last overload gave the following error.
    Type 'ZodObject<{ email: ZodString; password: ZodString; }, "strip", ZodTypeAny, { email: string; password: string; }, { email: string; password: string; }>' is not assignable to type 'ZodType<unknown, unknown, $ZodTypeInternals<unknown, unknown>> | ((intent: Intent | null) => ZodType<unknown, 
unknown, $ZodTypeInternals<unknown, unknown>>)'.
      Type 'ZodObject<{ email: ZodString; password: ZodString; }, "strip", ZodTypeAny, { email: string; password: string; }, { email: string; password: string; }>' is missing the following properties from type 'ZodType<unknown, unknown, $ZodTypeInternals<unknown, unknown>>': def, type, check, clone
, and 14 more.

14       return parseWithZod(formData, { schema });
                                         ~~~~~~

  node_modules/@conform-to/zod/dist/v3/parse.d.ts:23:5
    23     schema: Schema | ((intent: Intent | null) => Schema);
           ~~~~~~
    The expected type comes from property 'schema' which is declared here on type '{ schema: ZodType<unknown, unknown, $ZodTypeInternals<unknown, unknown>> | ((intent: Intent | null) => ZodType<unknown, unknown, $ZodTypeInternals<unknown, unknown>>); async: true; errorMap?: $ZodErrorMap<...> | unde
fined; formatError: (issues: $ZodIssue[]) => unknown; disableAutoCoercion?: boolean | undefined; }'
  node_modules/@conform-to/zod/dist/v3/parse.d.ts:22:25
    22 export declare function parseWithZod<Schema extends ZodTypeAny, FormError>(payload: FormData | URLSearchParams, options: {
                               ~~~~~~~~~~~~
    The last overload is declared here.


Found 1 error in app/routes/home.tsx:14
```

# Solution

Change all the import in v3 from "zod" to "zod/v3".
~~zod 3.21 has "zod/v3" import path so user don't have to update to zod 4~~ <- This is wrong. Described in discussion below

~~couldn't check my fix in my code because I was not able to publish or local install npm monorepo package... but all tests has passed!~~
WOW! pkg-pr-new made pkg for me! I installed it and error was gone! 
This workflow is incredible!

### DIFF
--- a/packages/conform-zod/package.json
+++ b/packages/conform-zod/package.json
@@ -4,16 +4,16 @@
 	"homepage": "https://conform.guide",
 	"license": "MIT",
 	"version": "1.13.3",
-	"main": "./dist/v3/index.js",
-	"module": "./dist/v3/index.mjs",
-	"types": "./dist/v3/index.d.ts",
+	"main": "./dist/default/index.js",
+	"module": "./dist/default/index.mjs",
+	"types": "./dist/default/index.d.ts",
 	"exports": {
 		".": {
-			"types": "./dist/v3/index.d.ts",
-			"module": "./dist/v3/index.mjs",
-			"import": "./dist/v3/index.mjs",
-			"require": "./dist/v3/index.js",
-			"default": "./dist/v3/index.mjs"
+			"types": "./dist/default/index.d.ts",
+			"module": "./dist/default/index.mjs",
+			"import": "./dist/default/index.mjs",
+			"require": "./dist/default/index.js",
+			"default": "./dist/default/index.mjs"
 		},
 		"./v3": {
 			"types": "./dist/v3/index.d.ts",
--- a/packages/conform-zod/rollup.config.js
+++ b/packages/conform-zod/rollup.config.js
@@ -14,6 +14,8 @@ function configurePackage() {
 			return !id.startsWith('.') && !path.isAbsolute(id);
 		},
 		input: [
+			`${sourceDir}/default/index.ts`,
+			`${sourceDir}/default/future.ts`,
 			`${sourceDir}/v3/index.ts`,
 			`${sourceDir}/v3/future.ts`,
 			`${sourceDir}/v4/index.ts`,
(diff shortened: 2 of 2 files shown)


edmundhung:
I think bumping the zod version in the peer dependency would be a breaking changes to people still using the older zod version :(

We can make it require zod v4 when we reach v2, which should happen at around Jan 2026.

chimame:
Just to be sure, is it correct that the definition of an import destination that supports Zod is as follows?

- `@conform-to/zod` supports Zod versions prior to v3.25.
- `@conform-to/zod/v3` supports the v3 method for Zod v3.25 and above.
- `@conform-to/zod/v4` supports the v4 method for Zod v3.25 and above.

I agree with this opinion. The only problem is that it's a little confusing that `/v3` is for use with Zod v4. 😅

edmundhung:
@kesoji Thanks for the quick update!

I wonder if it is more predictable if we duplicate the v3 folder and name it `default` directly. Just so we don't have to worry about the build script missing any edge cases.

I don't expect much changes on the packages and we can make sure we change both copies if needed.

edmundhung (approved):
Thank you both!

PR #1099, opened 2025-11-16
feat(conform-react/future): skip stale server result

### LINKED ISSUE #1095: Stale report `value` adds and removes characters during typing

🐛 **Bug summary**

Feature tests in https://github.com/edmundhung/conform/pull/1079 have surfaced a bug where the application of `targetValue` (see [`report`](https://conform.guide/api/react/future/report)) can lead to faulty form state.

The behavior presents when using `onChange={submitForm}` and having expensive renders following that. Within this timeframe, `targetValue` can become stale: The user has already typed ahead, which causes the application of `targetValue` to be destructive. In brief, input text gets overwritten as the user types.

**Expected:** As the user types, if a stale `targetValue` arrives from the server, the form update is discarded. User input does not compete with intents or reports.

**Actual:** As the user types, if a stale `targetValue` arrives from the server, conform updates the form with stale values. Some input characters are removed or added as the user is adding or deleting characters.

Quoting @edmundhung:

> This is gonna be really tricky to fix. Conform has no reliable way to know when a submission actually happens. We can detect when a submit event isn’t prevented, but that doesn’t cover submissions triggered with useSubmit(), which skips the event and reads form data directly from the DOM.

> That’s why I’ve been wondering if it’s reasonable to compare the current form value with the one the submission result was based on, to decide whether it’s still valid to use instead.

💻 **Reproduction demo**

To demonstrate this, the following setup features an `<input>` with `onInput={submitForm}` that renders 4,000 random numbers after every submission. This is to simulate a search page with a long result list.

<details>
<summary>
Code
</summary>

```ts
import { parseSubmission, report } from '@conform-to/dom/future';
import { useForm } from '@conform-to/react/future';
import { coerceFormValue } from '@conform-to/zod/v4/future';
import type { FormEventHandler } from 'react';
import { Form, useSubmit } from 'react-router';
import z from 'zod';

import type { Route } from './+types/home';

  z.object({
    input: z.string().optional(),
  })
);

export const loader = async ({ request }: Route.ActionArgs) => {
  const searchParams = new URL(request.url).searchParams;
  const submission = parseSubmission(searchParams);
  const result = schema.safeParse(submission.payload);

  if (!result.success) {
    throw result.error;
  }

  return {
    lastResult: result.data.input
      ? report(submission, {
          targetValue: {
            input: result.data.input,
          },
        })
      : null,
  };
};

export default ({ loaderData }: Route.ComponentProps) => {
  const { form, fields } = useForm({
    ...loaderData,
    schema,
  });

  const submit = useSubmit();

  const submitForm: FormEventHandler<HTMLInputElement> = (event) => {
    if (!event.currentTarget.form) {
      return;
    }

    submit(event.currentTarget.form, {
      replace: true,
    });
  };

  return (
    <Form {...form.props}>
      <input name={fields.input.name} onInput={submitForm} />
      <button type="submit">Submit</button>
      {new Array(4000).fill(0).map((_, i) => (
        <div key={i}>{Math.random()}</div>
      ))}
    </Form>
  );
};

```
</details>

[StackBlitz](https://stackblitz.com/edit/github-xql5392t?file=app%2Froutes%2Fhome.tsx)

### Conform version

v1.13.3

If you reproduce it locally, the following manual intervention is not necessary. But on StackBlitz, hit submit once before trusting the application to be properly initialized.

Once initialized, enter any text to observe the text mangling.

Others

### Screenshots or Videos

_No response_

### Additional context

_No response_

edmundhung:
Fixes #1095.

Conform now checks for stale server results to prevent overriding user changes. When a server result is received, it compares the current DOM state with the original submission payload. If the form was edited after submission (excluding fields skipped on the server or file inputs), the result is ignored.

This PR also renamed the `targetValue` option to `value`:

```tsx
// Update form to a specific value after submission
return {
  result: report(submission, {
    value: updatedData,
  }),
};

// Reset form to a specific value after submission
return {
  result: report(submission, {
    reset: true,
    value: updatedData,
  }),
};
```

### DIFF
--- a/packages/conform-react/future/hooks.tsx
+++ b/packages/conform-react/future/hooks.tsx
@@ -250,52 +250,49 @@ export function useConform<
 		[formRef, optionsRef],
 	);
 
+	// To avoid re-applying the same result twice
+	if (lastResult && lastResult !== lastResultRef.current) {
+		lastResultRef.current = lastResult;
+
+		const formElement = getFormElement(formRef);
+
+		if (
+			formElement &&
+			!isDirty(new FormData(formElement), {
+				defaultValue: lastResult.submission.payload,
+				intentName: optionsRef.current.intentName,
+				skipEntry(name) {
+					// Ignore fields the server intentionally excluded
+					return lastResult.submission.excludedFields.includes(name);
+				},
+				serialize(value) {
+					// Ignore files as server responses never include File objects
+					if (value instanceof File) {
+						return;
+					}
+
+					return optionsRef.current.serialize(value);
+				},
+			})
+		) {
+			handleSubmission('server', lastResult);
+		} else {
+			// FIXME: Remove debug log
+			// eslint-disable-next-line no-console
+			console.log('Skipping stale result', {
+				lastResult,
+				currentFormData: formElement ? getFormData(formElement) : null,
+				intentName: optionsRef.current.intentName,
+			});
+		}
+	}
+
 	useEffect(() => {
 		return () => {
 			// Cancel pending validation request
 			abortControllerRef.current?.abort('The component is unmounted');
 		};
 	}, []);
-
-	useEffect(() => {
-		// To avoid re-applying the same result twice
-		if (lastResult && lastResult !== lastResultRef.current) {
-			const formElement = getFormElement(formRef);
-
-			if (
-				!formElement ||
-				isDirty(new FormData(formElement), {
-					defaultValue: lastResult.submission.payload,
-					intentName: optionsRef.current.intentName,
-					skipEntry(name) {
-						// Ignore fields the server intentionally excluded
-						return lastResult.submission.excludedFields.includes(name);
-					},
-					serialize(value) {
-						// Ignore files as server responses never include File objects
-						if (value instanceof File) {
-							return;
-						}
-
-						return optionsRef.current.serialize(value);
-					},
-				})
-			) {
-				// FIXME: Remove debug log
-				// eslint-disable-next-line no-console
-				console.log('Skipping stale result', {
-					lastResult,
-					currentFormData: formElement ? getFormData(formElement) : null,
-					intentName: optionsRef.current.intentName,
-				});
-				// The form changed after submission. treat this result as stale.
-				return;
-			}
-
-			handleSubmission('server', lastResult);
-			lastResultRef.current = lastResult;
-		}
-	}, [lastResult, formRef, handleSubmission, optionsRef]);
 
 	useEffect(() => {
 		// Reset the form state if the form key changes


transparent-citizen:
Hope you could also get enough quality time for yourself throughout the weekend. The hard work you're putting in certainly doesn't go unnoticed. Thank you very much! 🙂

I'll give it a test and provide detailed feedback ASAP. Just wanted to share my appreciation right as I saw this PR!

edmundhung:
Thanks for the update @transparent-citizen!

Do you see the same issue with the pre-release on #1079? If not, I wonder if those form runs into the following cases:
- Forms with file input
- Forms with hidden inputs that are not managed by Conform (e.g. crsf token / honeypot)

transparent-citizen:
@edmundhung #1079 is my current baseline where I consider the reset behavior "good" (working as intended, minus the staleness bug). I compared these commits:

- [f358ea5](https://github.com/edmundhung/conform/pull/1079/commits/f358ea5c8361ca96f91f5a1a4876aae24256ca22) (from #1079), and
- [06adacb](https://github.com/edmundhung/conform/pull/1099/commits/06adacb13659dd1e97bc2ef1d1257ac6671ff2e2) (from this PR).

To narrow it down further:

- `useControl` isn't used anywhere
- No file inputs are present in the affected forms
- No `hidden` inputs are present in the affected forms

My hunch is that the freshness check of `lastResult` might not always be reliable.

From a theoretical standpoint, staleness is a concept closely related to time. So can a check for staleness truly be expressed without a notion of time? The current check compares form values, and it does fix the issue with quickly typing. But I wonder if the check breaks form updates in a subtle way.

Say you open a tin of cookies. As you look inside, you don't know when each individual cookie was baked. So there's no reliable way to tell which one is older and which is newer just by looking at them.

Isn't that what the current staleness check is doing though? (cookies = form states)

transparent-citizen:
Thanks for the debug setup! It helped me confirm that the affected forms break exactly when the debug message is printed. I've prepared [this repository](https://github.com/transparent-citizen/conform-staleness-bug) for you as a test case. There are two branches:

#### `master` ✅

The `master` branch is based on [f358ea5](https://github.com/edmundhung/conform/pull/1079/commits/f358ea5c8361ca96f91f5a1a4876aae24256ca22) which works as expected and acts as a baseline. There are two optional selects and a submit button. Selecting a meal will trigger a submission, which will then reset the form value to set an appropriate side dish.

- For example, try setting "Duck" as a meal, which should set "Onigiri" as a side dish. ✅
- Next, select "Ice cream" as a meal. The side dish updates to "Caramel syrup". ✅

image

#### `stale-result` ❌

The code here is identical, except that it uses this PR ([06adacb](https://github.com/edmundhung/conform/pull/1099/commits/06adacb13659dd1e97bc2ef1d1257ac6671ff2e2)). 

- First, select "Duck" as a meal, which should set "Onigiri" as a side dish. ✅
- Next, select "Ice cream" as a meal. This time, it does not update to "Caramel syrup". The console shows that conform erroneously rejects the update. ❌

image

- Lastly, if you manually press the submit button, the update succeeds. ✅

**Note**: As you switch to a new meal, the reset will be rejected if the previous reset was successful, so it keeps alternating between right and wrong behavior.

There is one exception: Switching between "Hamburger" and "Pasta" will always work. They use a common side dish ("Salad"). I'm not sure why this would change things, but since it's arcane behavior, I thought this paragraph might be helpful.

transparent-citizen:
Oh! Then the new render flow should be more efficient, not less efficient! In my experience, being able to drop a `useEffect` is typically a good sign. Very cool.

I reused the already familiar food example in a new [stale-result-redirect](https://github.com/transparent-citizen/conform-staleness-bug/tree/stale-result-redirect) branch (based on https://github.com/edmundhung/conform/commit/4d3a34ce1f6a6f4140fbeb8313fbd51c95d5a981), where I demonstrate another false positive. The code is very similar, but this time, note that it calls [`redirect`](https://reactrouter.com/dev/api/utils/redirect) several times in the loader.

The general idea of the redirects is to

- reduce the number of invalid `<select />` states, and to
- canonicalize valid `<select />` states.

For example, since the app knows the only possible side dish to a meal of duck to be onigiri, this redirect will occur:

```ini
                URL                  | Redirect rule
   --------------------------------- | -------------
   ?mealId=duck                      |
-> ?mealId=duck&sideDishId=onigiri   |      R3  
                ^^^^^^^^^^^^^^^^^^   |
```

Similarly, since the app knows not to offer ice cream with fried veggies, this redirect will occur:

```ini
                     URL                           Redirect rule
   --------------------------------------------- | -------------
   ?mealId=ice-cream&sideDishId=fried-veggies    |
-> ?mealId=ice-cream                             |        R2
-> ?mealId=ice-cream&sideDishId=caramel-syrup    |        R3
```

Here's an overview of the redirect rules that apply.

| Rule | Description |
| ---- | ----------- |
| R1    | Delete unrecognized `mealId` from URL |
| R2    | Delete `sideDishId` from URL if it doesn't belong to the meal |
| R3    | Set `sideDishId` in URL if there is only a single side dish available | 
| R4    | Delete `sideDishId` from URL if `mealId` is missing |

## Steps to reproduce

You can follow the same steps as in the other branches: Start in state A. Then set the meal (B). Right after the submission, conform will take us to state C and then log a console message about skipping a stale result. Like before, we never make it to state D.

#### State overview (copied from above)

| State | Meal   |     Side dish |
| -- | --------------|--------- |
| A | Duck |       Onigiri |
| B | Ice cream |  Onigiri |      
| C | Ice cream |  No selection |
| D| Ice cream |  Caramel syrup |

edmundhung:
> If memory serves me right, conform implicitly adds an intent somewhere in the submission. By the same token, could a client submission not add a state which starts at 0 and increments whenever the form submits

Only when you are using an intent (e.g. `intent.update()`), Conform will insert a button in your form and emulate a click on the button. So your form data will include additional entries. This is fine with POST form but could be quite awkward with GET form and they will be visible in the search params as well.

Most of the time, these intents are handled in the client. So it won't impact your actual URL. But if you need the server to handle an additional state, they have to be visible.

edmundhung:
Do you have another usecase with a POST form reporting a `value` from the server?

I wonder if the `value` option should focus on two usecases:
- Acts as the `defaultValue` for the form with intent button (progressive enhancement)
- Acts as the new `defaultValue` with `reset: true`

This way I will just rename the option to `defaultValue` to reflect that.

For GET form, we can try to provide a slightly different behavior when you provide a URLSearchParams as the default value of the `useForm()`:

```
function Example() {
  const { form, fields } = useForm({
    defaultValue: urlSearchParams,
  });

  // ...
}
```

I don't have a clear picture what the behavior difference would be but I think Conform should  support GET form better, include the redirect case.

edmundhung:
The reason why I wanna separate them is that updating a form value using a `redirect` makes more sense to me then having to specify a value in the `report` function.

Maybe if Conform has access to the URLSearchParams of the current URL, we could try to sync the values for you to support that?

edmundhung:
Decide to leave it as is for now as making it truly behaves like default value also comes with its own complexity 😅

PR #1112, opened 2025-12-02
feat: getFieldValue helper

edmundhung:
This PR introduces a `getFieldValue(formData, options)` helper to retrieve a field value from FormData with optional type guards:

```ts
import { getFieldValue } from '@conform-to/react/future';

// Basic: returns `unknown`
const email = getFieldValue(formData, 'email');

// With type guard: returns `string`, throws if not a string
const name = getFieldValue(formData, 'name', { type: 'string' });

// File type: returns `File`, throws if not a File
const avatar = getFieldValue(formData, 'avatar', { type: 'file' });

// Object type: parses nested fields into `{ city: unknown, ... }`
const address = getFieldValue<Address>(formData, 'address', { type: 'object' });

// Array: returns `unknown[]`
const tags = getFieldValue(formData, 'tags', { array: true });

// Array of objects: returns `Array<{ name: unknown, ... }>`
const items = getFieldValue<Item[]>(formData, 'items', {
  type: 'object',
  array: true,
});

// Optional: returns `string | undefined`, no error if missing
const bio = getFieldValue(formData, 'bio', { type: 'string', optional: true });
```

As it would be too complex to offer a type guard for object shape. All properties are mapped to the unknown type so you still get type hint when accessing individual properties. You can pass them to React and display the value just fine. Only if you have any specific logic based on the value, you will need to validate the property type yourself.

Here is a complete example:

```ts
import { useForm, useFormData, getFieldValue } from '@conform-to/react/future';

function Example() {
  const { form, fields } = useForm();
  // Retrieves the value of the `address` fieldset as an object, e.g. `{ city: unknown; ... }`
  const address = useFormData(form.id, (formData) =>
    getFieldValue(formData, fields.address.name, { type: 'object' }),
  );

  // ...
}
```

### DIFF
--- a/guide/app/layout.tsx
+++ b/guide/app/layout.tsx
@@ -61,6 +61,7 @@ const menus: { [code: string]: Menu[] } = {
 				{ title: 'useIntent', to: '/api/react/future/useIntent' },
 				{ title: 'parseSubmission', to: '/api/react/future/parseSubmission' },
 				{ title: 'report', to: '/api/react/future/report' },
+				{ title: 'getFieldValue', to: '/api/react/future/getFieldValue' },
 				{ title: 'isDirty', to: '/api/react/future/isDirty' },
 				{ title: 'FormProvider', to: '/api/react/future/FormProvider' },
 				{


oxc:
I like it, looks safe, covers many basic use cases, and the implementation still seems maintainable. 

Instinctively I would say why not also add support for `{ type: "object", schema: standardSchema }`, but that would certainly complicate things, and can be added rather simple as a custom helper.

edmundhung:
Released in [v1.15.0](https://github.com/edmundhung/conform/releases/tag/v1.15.0).

PR #1126, opened 2025-12-16
fix(conform-to/zod) add support for nullable constraints in getZodConstraint

a1r93:
### Bug details

Take for example the following zod schema:

```ts
const schema = z.object({
    value: z.number().min(0).max(100).nullable(),
})
```

When calling `getZodConstraint`, the output is `{ required: false }`, the min and max constraints are omited.

### Fix

Add a condition for the `nullable` type and handle it in the same way `optional` is handled.

### DIFF
--- a/packages/conform-zod/v4/tests/constraint.test.ts
+++ b/packages/conform-zod/v4/tests/constraint.test.ts
@@ -83,7 +83,7 @@ describe('constraint', () => {
 				min: 1,
 				max: 10,
 			},
-			nullishNumber: {
+			nullableNumber: {
 				required: true,
 				min: 1,
 				max: 10,


packages/conform-zod/v4/constraint.ts
```diff
@@ -1,6 +1,6 @@
 import type { Constraint } from '@conform-to/dom';
 
-import { $ZodType, $ZodTypes, $ZodNumber, $ZodString } from 'zod/v4/core';
+import { $ZodNumber, $ZodString, $ZodType, $ZodTypes } from 'zod/v4/core';
```
chimame: It is best not to make any changes that are different from the fix.

packages/conform-zod/v4/constraint.ts
```diff
@@ -106,6 +106,9 @@ export function getZodConstraint(schema: $ZodType): Record<string, Constraint> {
 		} else if (def.type === 'optional') {
 			constraint.required = false;
 			updateConstraint(def.innerType, data, name);
+		} else if (def.type === 'nullable') {
+			constraint.required = false;
```
chimame: `nullable` allows null values, it doesn't make them optional, so this line shouldn't be necessary.

PR #1131, opened 2025-12-20
feat: introduce `configureForms` API

edmundhung:
Close #1069

`configureForms` is a factory function that creates customized form hooks. It replaces the `FormOptionsProvider` by giving you a set of customized React hooks with scoped types based on the configuration:

```ts
import { configureForms } from '@conform-to/react/future';

const { FormProvider, useForm, useFormMetadata, useField, useIntent } =
  configureForms({
    // configs...
  });
```

The current implementation supports the following customization:

- Extracting HTML constraints from schema:

  ```ts
  // For Zod v3
  import { getConstraints } from '@conform-to/zod/v3/future';

  // For Zod v4
  import { getConstraints } from '@conform-to/zod/v4/future';

  // For Valibot
  import { getConstraints } from '@conform-to/valibot/future';
  ```

  ```ts
  import { configureForms } from '@conform-to/react/future';
  import { getConstraints } from '@conform-to/zod/v3/future';

  const { useForm } = configureForms({
    getConstraints,
  });
  ```
- Customizing schema validation

  ```ts
  import { configureForms } from '@conform-to/react/future';
  import { formatResult } from '@conform-to/zod/v3/future';
  import type { ZodTypeAny, input, output } from 'zod';
  import { z } from 'zod';

  declare module '@conform-to/react/future' {
    interface CustomSchemaTypes<Schema> {
      input: Schema extends ZodTypeAny ? input<Schema> : never;
      output: Schema extends ZodTypeAny ? output<Schema> : never;
      options: Schema extends ZodTypeAny ? { errorMap?: z.ZodErrorMap } : never;
    }
  }

  const { useForm } = configureForms({
    isSchema(schema): schema is z.ZodTypeAny {
      return schema instanceof z.ZodType;
    },
    validateSchema(schema, payload, options) {
      try {
        const result = schema.safeParse(payload, options);
        return formatResult(result);
      } catch {
        return schema.safeParseAsync(payload, options).then(formatResult);
      }
    },
  });
  ```

- Integrating with UI libraries

  ```ts
  import { configureForms, shape } from '@conform-to/react/future';

  const { useForm } = configureForms({
    // Restrict the ErrorShape to type "string"
    isError: shape<string>(),
    extendFormMetadata(metadata) {
      return {
        get props() {
          return {
            ...metadata.props,
            isInvalid: !metadata.valid,
            errors: metadata.errors,
          } satisfies Partial<React.ComponentProps<typeof Form>>;
        },
      };
    },
    extendFieldMetadata(metadata, { when }) {
      return {
        // Available on all fields
        get textFieldProps() {
          return {
            name: metadata.name,
            defaultValue: metadata.defaultValue,
            required: metadata.required,
            isInvalid: !metadata.valid,
            errors: metadata.errors,
          } satisfies Partial<React.ComponentProps<typeof TextField>>;
        },
        // Only available for fields with a specific shape
        get dateRangePickerProps() {
          return when(
            metadata,
            shape<{ start: string; end: string }>(),
            ({ valid, errors, getFieldset }) => {
              const rangeFields = getFieldset();

              return {
                startName: rangeFields.start.name,
                endName: rangeFields.end.name,
                defaultValue: {
                  start: rangeFields.start.defaultValue,
                  end: rangeFields.end.defaultValue,
                },
                isInvalid: !valid,
                errors,
              } satisfies Partial<React.ComponentProps<typeof DateRangePicker>>;
            },
          );
        },
      };
    },
  });
  ```

### DIFF
--- a/packages/conform-react/future/types.ts
+++ b/packages/conform-react/future/types.ts
@@ -809,6 +809,9 @@ export type BaseMetadata<
 	ErrorShape extends BaseErrorShape,
 > = BaseFieldMetadata<FieldShape, ErrorShape>;
 
+/**
+ * @deprecated Use `configureForms()` with the `extendFieldMetadata` option instead.
+ */
 export type CustomMetadataDefinition = <
 	FieldShape,
 	ErrorShape extends BaseErrorShape,


packages/conform-react/future/state.ts
```diff
@@ -502,45 +542,20 @@ export function getField<FieldShape, ErrorShape = string>(
...
-
-			throw new Error(
-				`Property "${String(prop)}" does not exist on field metadata. ` +
-					`If you have defined the CustomMetadata interface to include "${String(prop)}", make sure to also implement it through the "defineCustomMetadata" property on <FormOptionsProvider />.`,
-			);
-		},
-	});
+	return Object.assign(
+		metadata,
+		customize?.(metadata, { form, when }),
```
uncrft: Is there a particular reason why `when` is available in the context but `shape` isn't? 

Is it because `shape` can be used for `isError`, or even `isSchema` with a runtime type guard, while `when` can only used for `extendFieldMetadata`? 

I'm trying to think of other use cases for the `when` function that could justify an export, but I can't think of anything at the factory level. Maybe at the hook level?

packages/conform-react/future/util.ts
```diff
@@ -267,3 +272,26 @@ export function compactMap<Item>(
 export function generateUniqueKey() {
 	return Math.trunc(Date.now() * Math.random()).toString(36);
 }
+
+/**
+ * Creates a type guard function for specifying field shape constraints.
+ */
+export function shape<T>(
```
oxc: I find the `shape` helper a bit misleading. Without an actual guard implementation, it acts like an unchecked cast, but still looks like it provides some safety.

I know the API design looks pretty like this, but I personally don't like implicit `any` casts.

What about having a `shape<T>(guard)` and a `assumeShape<T>()` or something like this?

packages/conform-react/future/state.ts
edmundhung: > Is it because shape can be used for isError, or even isSchema with a runtime type guard, while when can only used for extendFieldMetadata?

Correct. I had both exported at first. But it's unlikely to reuse the `when` helper in other context, while `shape` might still be useful elsewhere.

I have also thought of making `configureForms` provide the `shape` helper so you don't need to import it additionally, but that would likely means making configureForms accepting a callback function instead of an object like now.

packages/conform-react/future/util.ts
edmundhung: Haha, there are a lot of types ~~magic~~ in the API to achieve the type inference behavior I wanted.

`shape` is currently used in 2 places: `isError` config, or to use with the `when` helper. They are both declared as a type guard but we are not actually using them at runtime at all.

There are a few reasons, but the main one is to allow providing custom types while the rest of the types remains inferred. These types are used as a constraint in the type system. For example, the shape you provide to `when` is used to keep a marker in the custom field metadata, so we can make them conditionally available based on the shape of the field. As Conform has no idea of the field shape at runtime anyway, it's simply used to carry the types over.

As the "safety" here is purely at the type level, how about simplifying `shape` to accept no arguments and always return `true` to keep it honest?

```ts
export function shape<T>(): (value: unknown) => value is T {
  return (): value is T => true;
}
```

packages/conform-react/future/util.ts
```diff
@@ -267,3 +275,53 @@ export function compactMap<Item>(
...
+		schema !== null &&
+		'~standard' in schema &&
+		typeof schema['~standard'] === 'object' &&
+		schema['~standard'] !== null &&
+		'version' in schema['~standard'] &&
+		schema['~standard'].version === 1
+	);
+}
+
+export function valdiateStandardSchemaV1<Schema extends StandardSchemaV1>(
```
uncrft:
```suggestion
export function validateStandardSchemaV1<Schema extends StandardSchemaV1>(
```

transparent-citizen:
This looks exciting. The way this previously worked via module augmentation wasn't very idiomatic and modeled type assertion rather than inference. It worked really well though!

```js
// Export helper types
export type FormMetadata = InferFormMetadata<typeof result.config>;
export type FieldMetadata<FieldShape> = InferFieldMetadata<
  typeof result.config,
  FieldShape
>;
```

I've had trouble getting these to infer properly. I tested them with

```js
"@conform-to/react": "https://pkg.pr.new/@conform-to/react@b0bfec8",
"@conform-to/zod": "https://pkg.pr.new/@conform-to/zod@b0bfec8",
```

but each evaluates to `never`. Are their definitions still up to date?

edmundhung:
That's an oversight from me.

Instead of InferFormMetadata and InferFieldMetadata returning full types, I wonder having some type helpers that return just the custom parts would be more flexible as you can compose them like this:

```ts
import {
  type FormMetadata as BaseFormMetadata,
  type FieldMetadata as BaseFieldMetadata,
  type Fieldset as BaseFieldset,
  type InferBaseErrorShape,
  type InferCustomFormMetadata,
  type InferCustomFieldMetadata,
} from '@conform-to/react/future';

const { config } = configureForms({ ... });

type BaseErrorShape = InferBaseErrorShape<typeof config>;
type CustomFormMetadata = InferCustomFormMetadata<typeof config>;
type CustomFieldMetadata = InferCustomFieldMetadata<typeof config>;

export type FormMetadata<ErrorShape extends BaseErrorShape = BaseErrorShape> = BaseFormMetadata<ErrorShape, CustomFormMetadata, CustomFieldMetadata>;
export type FieldMetadata<FieldShape, ErrorShape extends BaseErrorShape = BaseErrorShape> = BaseFieldMetadata<FieldShape, ErrorShape, CustomFieldMetadata>;
export type Fieldset<FieldShape, ErrorShape extends BaseErrorShape = BaseErrorShape> = BaseFieldset<FieldShape, ErrorShape, CustomFieldMetadata>;
```

This might be a bit more messy to write though, or would you prefer the current approach and I should just provide a `InferFieldset` helper type?

transparent-citizen:
React moved from CSR to SSR to RSC, and with each new render model, JS execution on the client was reduced further. Frameworks like react-router are picking this mantra up, preaching to prefer using JS for progressive enhancement rather than building core functionality to save you from ending up with a site that's broken for the `<noscript>` segment of your users.

At the same time, browser vendors have been shipping new HTML elements to enable building accordions, modals, and recently anchored popovers with HTML and CSS only. As availability goes up, future UI frameworks can be expected to have to work under stricter conditions, so there's an incentive to keep client-side JS minimal. That said, among all the popular UI libraries in the React ecosystem, there currently isn't a single one that's usable with JS disabled, not even a "simple" MUI TextField.

I understand that Conform must be designed for today's world. But treating elements that are native to the web as first-class citizens seems like a pretty safe bet to me. `<input>` will still be around in a decade, along with its attributes and 10 years' worth of upcoming CSS development. But a MUI Date Picker will eventually just fade away (or turn to CSS).

In the meantime, couldn't we have both? Metadata getters for native HTML elements as well as a more generalized way for UI library scaffolding?

PR #1134, opened 2025-12-27
docs: add Ja translation in `useControl.md`

WataruNishimura:
# Overview
- It seems that there are no issues or discussions about japanese translation of useControl.md.
- I translated doc and made this PR by myself(not genAI).
# Remarks
- This is a literal translation. ( like other ja docs )
- Some words ( e.g. hidden) are used without translation.


WataruNishimura:
I think Japanese documentation is necessary, but I don't really mind how we make or translate them.

I'd like more Japanese developers to use this library, because I believe Conform is one of the best form library for Next.js or other React meta frameworks.
In that case, I thought it'd be great if there were a systematically organised natural Japanese document that's easy for Japanese people to grasp at a glance.


1. I thought now is not the time to create Japanese`useControl.md` document.
2. It'll be great if there were Japanese documentations.

PR #1147, opened 2026-01-23
feat: add `preserveOptionsOrder` option to useControl for multi-select

edmundhung:
Fix https://github.com/edmundhung/conform/discussions/1144

### DIFF
--- a/.changeset/preserve-options-order.md
+++ b/.changeset/preserve-options-order.md
@@ -1,0 +1,17 @@
+---
+'@conform-to/dom': minor
+'@conform-to/react': minor
+---
+
+Add `preserveOptionsOrder` option to `useControl` for multi-select inputs.
+
+By default, `<select multiple>` returns selected values in DOM order. With `preserveOptionsOrder: true`, the order passed to `control.change()` is preserved, and form reset restores the `defaultValue` order.
+
+```tsx
+const control = useControl({
+  defaultValue: ['c', 'a', 'b'],
+  preserveOptionsOrder: true,
+});
+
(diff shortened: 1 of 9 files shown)


oxc:
Unfortunately this creates duplicates in my `<select multiple>` which seem to break my autocomplete, and I have not been able to figure out why exactly.
I guess since you are not pursuing this approach anyway, I will just wait for your generic version.

edmundhung:
That sounds like a bug on how this PR manages the option element in the DOM. I will start working on the updated `useControl` API this week now that I have the [coerceStructure()](https://github.com/edmundhung/conform/pull/1169) API ready to get the type safety I want 😄

oxc:
> Unfortunately this creates duplicates in my `<select multiple>` which seem to break my autocomplete, and I have not been able to figure out why exactly.

This seems to have been caused by me adding the field name to both the input field and the hidden select 🤦🏻‍♂️

oxc:
Do I understand correctly that I should now be able to implement this with the updated useControl hook controlling a fieldset with single hidden inputs for each array value?

edmundhung:
@oxc yes, you should be able to use a `<BaseControl type="fieldset" />` to achieve what you want. It should work with any shape, including a string array. Just let me know if there is anything unclear and I can look into it with you.

oxc:
No, it works nicely, and I can move the conversion between strings and objects directly into the useControl hook 👍🏻 Thanks!

edmundhung:
Nice! I am gonna close this then. 👍🏼

PR #1175, opened 2026-03-14
Support JSON fields via `nameForJSON` and `defaultJSON`

lhapaipai:
**Closes #1137**

### Context

When using `coerceFormValue` / `configureCoercion` with custom object types (e.g. a media picker), developers face two friction points today:

1. **Data preparation** — the `defaultValue` passed to `useForm` must already be serialized as a string, requiring a manual `JSON.stringify` before passing loader data to the form.
2. **TypeScript error** — since the schema expects an object (e.g. `Media`) but the form receives a string, TypeScript rejects the `defaultValue` assignment.

This was surfaced in [#1135](https://github.com/edmundhung/conform/discussions/1135) and you suggested extending Conform's naming conventions to support automatic JSON serialisation via a dedicated name pattern.

### Solution

This PR implements that suggestion by adding two new fields to `FieldMetadata`:

- **`nameForJSON`** — same name as `name`, typed as `FieldName<string>` instead of `FieldName<FieldShape>`. Use it as the `name` attribute on your input when the value is a JSON-serialised object.
- **`defaultJSON`** — the field's default value serialised via `JSON.stringify`. Use it as the `defaultValue` attribute.

```tsx
<input
  name={fields.media.nameForJSON}
  defaultValue={fields.media.defaultJSON}
/>
```

This pair allows you to pass the original typed object directly to `defaultValue` in `useForm` no manual stringification needed. while still submitting the value as a flat JSON string in the `FormData`.

### Example

See `examples/react-router/app/routes/json.tsx` for a full working demo with a `mediaSchema`.

---

### Open question — dirty detection

The current example uses `isDirty` with a custom `serialize` option to compare the form's current state against the default value:

```tsx
const dirty = useFormData(
  form.id,
  (formData) =>
    isDirty(formData, {
      defaultValue: form.defaultValue,
      serialize: (value, defaultSerialize) => {
        if (
          value instanceof Object &&
          'type' in value &&
          value.type === 'media'
        ) {
          return JSON.stringify(value);
        }
        return defaultSerialize(value);
      },
    }) ?? false,
);
```

This works but feels **fragile**: the detection relies on duck-typing (`value.type === 'media'`) rather than on the schema definition itself. As soon as there are multiple custom JSON types, this serialize function grows and becomes a maintenance burden, with no guarantee that it stays in sync with the `customize` coercion config.

Happy to hear if there's a better direction for this, or if the current duck-typing approach is the intended pattern for now.

<details>

- Adds FieldMetadata.nameForJSON: FieldName<string> and FieldMetadata.defaultJSON: string (getter) in packages/conform-react/future/types.ts and initializes them in packages/conform-react/future/state.ts.
- Adds export getDefaultJSON(context, name) in packages/conform-react/future/state.ts which resolves the field's server/target/default value and returns JSON.stringify(value) (falls back to '').
- Example route examples/react-router/app/routes/json.tsx demonstrating Zod schema + custom coercion for media objects, usage of fields.media.nameForJSON and fields.media.defaultJSON in inputs, and a custom serialize-based dirty check.
- Adds navigation and route config entries to expose the new example route.

Example usage:
```jsx
<input name={fields.media.nameForJSON} defaultValue={fields.media.defaultJSON} />
```

</details>

### DIFF
--- a/examples/react-router/app/root.tsx
+++ b/examples/react-router/app/root.tsx
@@ -58,6 +58,9 @@ export default function App() {
 				<li>
 					<Link to="todos">Dynamic form with data persistence</Link>
 				</li>
+				<li>
+					<Link to="date">Json</Link>
+				</li>
 			</ul>
 
 			<hr />
--- a/examples/react-router/app/root.tsx
+++ b/examples/react-router/app/root.tsx
@@ -59,7 +59,7 @@ export default function App() {
 					<Link to="todos">Dynamic form with data persistence</Link>
 				</li>
 				<li>
-					<Link to="date">Json</Link>
+					<Link to="json">Json</Link>
 				</li>
 			</ul>
 
--- a/examples/react-router/app/routes.ts
+++ b/examples/react-router/app/routes.ts
@@ -8,5 +8,6 @@ export default [
 	route('signup-async-schema', 'routes/signup-async-schema.tsx'),
 	route('signup-server-validation', 'routes/signup-server-validation.tsx'),
 	route('todos', 'routes/todos.tsx'),
+	route('json', 'routes/json.tsx'),
 	route('file-upload', 'routes/file-upload.tsx'),
 ] satisfies RouteConfig;
--- a/examples/react-router/app/routes/json.tsx
+++ b/examples/react-router/app/routes/json.tsx
@@ -21,6 +21,10 @@ const userSchema = z.object({
 	media: mediaSchema,
 });
 
+// not ideal, performance issues ?
+const isMediaValue = (value: unknown): value is z.infer<typeof mediaSchema> =>
+	mediaSchema.safeParse(value).success;
+
 const { coerceFormValue } = configureCoercion({
 	customize(type) {
 		if (type === mediaSchema) {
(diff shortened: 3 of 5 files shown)


edmundhung:
Hey @lhapaipai, thanks for giving it a spin!

I haven't looked into your implementation yet. But I have been working on #1173 that might be a solution for your needs too. I am still finalizing some details. But the idea is to let you register a complex field through a `<fieldset>` element. It will provide you a component that will render the necessary inputs to represent your data and you can change / focus / blur on it like any regular inputs:

```tsx
import { HiddenInput, useControl } from '@conform-to/react/future';

interface DateRange {
  start: string;
  end: string;
}

function DateRangeField(props: { name: string; defaultValue: DateRange }) {
  const control = useControl({
    defaultValue: props.defaultValue,
    parse(payload) {
      return DateRangeSchema.parse(payload);
    },
  });

  return (
    <>
      <HiddenInput
        type="fieldset"
        name={props.name}
        ref={control.register}
        defaultValue={control.defaultValue}
      />
      <CustomDateRangePicker
        value={control.payload}
        onChange={(value) => control.change(value)}
        onBlur={() => control.blur()}
      />
    </>
  );
}
```

In this case, the `HiddenInput` component will render two inputs for you automatically
- `<input name={props.name + '.start'} defaultValue={control.defaultValue.start} />`
- `<input name={props.name + '.end'} defaultValue={control.defaultValue.end} />`

This also works with more complex structure like an array of objects. There are some challenge on the type safety as well but I believe I am mostly there. I am hoping to have something more reliable by the end of this weekend, then I will invite some of you for a test. 👍🏼

lhapaipai:
Very interesting, it seems it can be another approach that solves the dirty detection issue.

lhapaipai:
#1180 provides a different implementation that solves the same issue

PR #1188, opened 2026-04-16
feat: set `pattern` constraint for regular expressions

transparent-citizen:
This PR lets Conform construct an HTML [`pattern`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/pattern) attribute based on regular expressions from zod v3, v4, and valibot schemas.

These validators allow multiple regexes for string schemas. For example in zod,

```ts
z.string().regex(/a/).regex(/b/)
```

matches strings containing at least one "a" and one "b" in any order. But because `<input>` accepts only one `pattern` attribute, multiple regexes need to be combined into one, using positive lookahead assertions:

```py
 in: [A-Z], [0-9]
out: ^(?=.*(?:[A-Z]))(?=.*(?:[0-9])).*$
```

The correctness of `out` is supported by [this jsfiddle](https://jsfiddle.net/8co0aLjw/) featuring an input whose text turns green if and only if it matches both `in` patterns.

The `i` flag is not supported, because HTML patterns don't support case-insensitive matching.

Backreferences have limited support. Using named backreferences over numbered backreferences is recommended.

<details>

- Added combinePatterns(patterns: RegExp[]): string | undefined in packages/conform-dom/util.ts and re-exported it from packages/conform-dom/future. It returns undefined for empty input or when any regex has unsupported flags (allowed flags u/v are stripped). For valid inputs it builds a single anchored pattern that conjuncts multiple regexes with positive lookaheads (e.g. ^(?=.*(?:[A-Z]))(?=.*(?:[0-9])).*$). HTML pattern generation assumes browser behavior using the v flag by default; supported flags are stripped and unsupported flags are not mapped.  
- Zod (default): getZodConstraint now collects ZodString regex checks (def.checks kind === 'regex') and sets constraint.pattern via combinePatterns.  
- Zod v4: getZodConstraint now reads _schema._zod.bag.patterns and sets constraint.pattern via combinePatterns.  
- Valibot: getValibotConstraint now inspects schema.pipe for regex validators and sets constraint.pattern via combinePatterns.  
- Tests: Added unit tests for combinePatterns and Vitest cases verifying single and chained regex → combined pattern behavior for zod/default, zod/v4, and valibot; tests cover empty-input and unsupported/supported flags.  
- Changeset: Adds documentation and example showing mapping of schema regex validators to the HTML pattern attribute and how to expose it via inputProps.

</details>

### DIFF
--- a/packages/conform-dom/tests/util.test.ts
+++ b/packages/conform-dom/tests/util.test.ts
@@ -1,5 +1,5 @@
 import { expect, test } from 'vitest';
-import { serializeHtmlPattern, deepEqual } from '../util';
+import { deepEqual, serializeHtmlPattern } from '../util';
 
 test('deepEqual', () => {
 	expect(deepEqual(null, null)).toBe(true);
@@ -41,13 +41,16 @@ test('serializeHtmlPattern', () => {
 	);
 
 	// Unsupported flags
-	expect(serializeHtmlPattern([/[A-Z]/g])).toBeUndefined();
+	expect(serializeHtmlPattern([/[A-Z]/i])).toBeUndefined();
 	expect(serializeHtmlPattern([/[A-Z]/, /[0-9]/i])).toBeUndefined();
 
-	// Supported flags
+	// Other flags
+	expect(serializeHtmlPattern([/[A-Z]/g])).toBe('^(?=.*(?:[A-Z])).*$');
+	expect(serializeHtmlPattern([/[A-Z]/m])).toBe('^(?=.*(?:[A-Z])).*$');
+	expect(serializeHtmlPattern([/[A-Z]/s])).toBe('^(?=.*(?:[A-Z])).*$');
 	expect(serializeHtmlPattern([/[A-Z]/u])).toBe('^(?=.*(?:[A-Z])).*$');
 
-	// Semantic validation: single pattern (uppercase-only)
+	// Semantic validation: single pattern (uppercase)
 	const single = serializeHtmlPattern([/^[A-Z]+$/]);
 	expect(single).toBeDefined();
 	const singleRegex = new RegExp(single || '');
@@ -55,7 +58,7 @@ test('serializeHtmlPattern', () => {
 	expect(singleRegex.test('abc')).toBe(false);
 	expect(singleRegex.test('ABC123')).toBe(false);
 
-	// Semantic validation: multiple patterns (uppercase AND digit)
+	// Semantic validation: multiple patterns (uppercase and digit)
 	const multi = serializeHtmlPattern([/[A-Z]/, /[0-9]/]);
 	expect(multi).toBeDefined();
 	const multiRegex = new RegExp(multi || '');
--- a/packages/conform-dom/tests/util.test.ts
+++ b/packages/conform-dom/tests/util.test.ts
@@ -1,5 +1,5 @@
 import { expect, test } from 'vitest';
-import { deepEqual } from '../util';
+import { combinePatterns, deepEqual } from '../util';
 
 test('deepEqual', () => {
 	expect(deepEqual(null, null)).toBe(true);
@@ -21,4 +21,42 @@ test('deepEqual', () => {
 	expect(deepEqual(new File([], 'example'), new File([], 'example'))).toBe(
 		false,
 	);
+});
+
+test('combinePatterns', () => {
+	// Empty pattern
+	expect(combinePatterns([])).toBeUndefined();
+
+	// Single pattern
+	expect(combinePatterns([/[A-Z]/])).toBe('^(?=.*(?:[A-Z])).*$');
+	expect(combinePatterns([/^[A-Z]+$/])).toBe('^(?=.*(?:^[A-Z]+$)).*$');
+
+	// Multiple patterns without flags
+	expect(combinePatterns([/[A-Z]/, /[0-9]/])).toBe(
+		'^(?=.*(?:[A-Z]))(?=.*(?:[0-9])).*$',
+	);
+
+	// Unsupported flags
+	expect(combinePatterns([/[A-Z]/g])).toBeUndefined();
+	expect(combinePatterns([/[A-Z]/, /[0-9]/i])).toBeUndefined();
+
+	// Supported flags
+	expect(combinePatterns([/[A-Z]/u])).toBe('^(?=.*(?:[A-Z])).*$');
+
+	// Semantic validation: single pattern (uppercase-only)
+	const single = combinePatterns([/^[A-Z]+$/]);
+	expect(single).toBeDefined();
+	const singleRegex = new RegExp(single || '');
+	expect(singleRegex.test('ABC')).toBe(true);
+	expect(singleRegex.test('abc')).toBe(false);
+	expect(singleRegex.test('ABC123')).toBe(false);
+
+	// Semantic validation: multiple patterns (uppercase AND digit)
+	const multi = combinePatterns([/[A-Z]/, /[0-9]/]);
+	expect(multi).toBeDefined();
+	const multiRegex = new RegExp(multi || '');
+	expect(multiRegex.test('ABC')).toBe(false);
+	expect(multiRegex.test('123')).toBe(false);
+	expect(multiRegex.test('ABC123')).toBe(true);
+	expect(multiRegex.test('abc123')).toBe(false);
 });
--- a/packages/conform-dom/tests/util.test.ts
+++ b/packages/conform-dom/tests/util.test.ts
@@ -1,5 +1,5 @@
 import { expect, test } from 'vitest';
-import { combinePatterns, deepEqual } from '../util';
+import { serializeHtmlPattern, deepEqual } from '../util';
 
 test('deepEqual', () => {
 	expect(deepEqual(null, null)).toBe(true);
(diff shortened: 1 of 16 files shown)


packages/conform-dom/util.ts
```diff
@@ -113,3 +113,23 @@ export function getTypeName(value: unknown): string {
...
+	if (patterns.length === 0) {
+		return '';
+	}
+
+	const first = patterns[0];
+	if (first && patterns.length === 1) {
+		return first.source;
+	}
+
+	return `^${patterns.map((p) => `(?=.*${p.source})`).join('')}.*$`;
```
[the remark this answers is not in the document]
transparent-citizen: Your proposed fix would explain the first 4 lines. But why do we need to put a non-capturing group inside the lookahead assertion?

packages/conform-dom/tests/util.test.ts
```diff
@@ -22,3 +22,21 @@ test('deepEqual', () => {
...
+
+	// Multiple patterns without flags
+	expect(combinePatterns([/[A-Z]/, /[0-9]/])).toBe(
+		'^(?=.*(?:[A-Z]))(?=.*(?:[0-9])).*$',
+	);
+
+	// Patterns with unsupported flags
+	expect(combinePatterns([/[A-Z]/g])).toBeUndefined();
+	expect(combinePatterns([/[A-Z]/, /[0-9]/i])).toBeUndefined();
+});
```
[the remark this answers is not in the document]
transparent-citizen: I added a test for the `u` flag, but chose to pass on the `v` flag. The syntax is relatively new, and requiring newer syntax might limit environment support.

@edmundhung: Willing to change this if you prefer.

transparent-citizen:
@edmundhung Thanks for setting up CodeRabbit! A lot of its suggestions were quite helpful. Looking forward to hearing from you! 🙂

packages/conform-dom/util.ts
```diff
@@ -113,3 +113,25 @@ export function getTypeName(value: unknown): string {
...
+ * Combines multiple regex patterns into a single HTML pattern attribute.
+ *
+ * This is done by putting each input pattern into a lookahead assertion.
+ *
+ * HTML pattern attributes implicitly run with the `v` flag, so `u` and `v` flags are
+ * silently stripped. Returns undefined if any pattern has other flags (e.g., `g`, `i`).
+ *
+ * Example: [/[A-Z]/, /[0-9]/] -> '^(?=.*(?:[A-Z]))(?=.*(?:[0-9])).*$'
+ */
+export function combinePatterns(patterns: RegExp[]): string | undefined {
```
edmundhung: How about naming it `serializeHtmlPattern(pattern: RegExp | RegExp[])`? As I consider it a generic utility to transform one or more regex to a html pattern string.

packages/conform-dom/util.ts
```diff
@@ -113,3 +113,25 @@ export function getTypeName(value: unknown): string {
...
+	if (patterns.length === 0) {
+		return undefined;
+	}
+
+	if (patterns.some((p) => p.flags.replace(/[uv]/g, ''))) {
+		return undefined;
+	}
```
edmundhung: I would suggest making it best-effort, so remaining valid regex should still be used to generate a HTML pattern string.

I also wonder if we should validate the regex manually using the `v` flag, like this:

```
try {
  new RegExp(pattern, 'v');
  return true;
} catch {
  return false;
}
```

What do you think?

packages/conform-valibot/constraint.ts
```diff
@@ -149,6 +150,22 @@ export function getValibotConstraint<
...
+			// @ts-expect-error
+			const regexValidators = schema.pipe?.filter(
+				// @ts-expect-error
+				(v) => 'type' in v && v.type === 'regex',
+			);
+			if (regexValidators?.length) {
+				const pattern = combinePatterns(
+					regexValidators.map(
+						// @ts-expect-error
+						(v) => (v as { requirement: RegExp }).requirement,
```
edmundhung: Nitpick: Since we are bypassing type checking here anyway, there is no need to cast the type.

```suggestion
						(v) => v.requirement,
```

packages/conform-zod/default/constraint.ts
```diff
@@ -129,6 +131,16 @@ export function getZodConstraint(
...
+			const regexChecks = def.checks.filter(
+				(check): check is ZodStringCheck & { kind: 'regex' } =>
+					check.kind === 'regex',
+			);
+			if (regexChecks.length > 0) {
+				const pattern = combinePatterns(regexChecks.map((c) => c.regex));
+				if (pattern) {
+					constraint.pattern = pattern;
+				}
+			}
```
edmundhung, line 143: You will need to make the same changes and tests in the `v3` directory. The difference comparing to the `default` directory is that `v3` imports from `zod/v3`, so it works for v4 users who still  uses a v3 schema somewhere, while `default` only works with v3 versions.

We will get rid of the default directory in the next major.

edmundhung (commented):
Love it! Thanks for working on this 👍🏼

packages/conform-dom/util.ts
transparent-citizen: I looked into the unsupported flags a bit more. From what I can tell, we can safely drop all of them without reducing the validity of the pattern, the only exception being `i`. 

However, `/a/i` and `/[aA]/` describe the same grammar, and similarly we can map `/[a-z]/i` to `/[a-zA-Z]/` to make the regex flagless. It looks feasible to map all `i`-flagged regexes to flagless regexes.

I think validity checks like the one you showed are reasonable to do, but I think tests are a more suitable environment for this code. Runtime verification makes Conform pay a runtime cost, as it would load a new `RegExp` object into memory and compile the pattern to a finite-state machine, when all it needed to do was to combine the input patterns. 

Is it at all possible to enter the `catch` case? Given that the inputs to `serializeHtmlPattern` are `RegExp` objects with valid patterns, and we're placing them in isolated slots using `(?:...)`, it couldn't possibly yield an unparseable regex, can it?

packages/conform-zod/default/constraint.ts
transparent-citizen: Thanks for the explanation! Fixed via https://github.com/edmundhung/conform/pull/1188/changes/b59535ef58b216801b283fc67d4ba84db5df8367.

packages/conform-valibot/constraint.ts
transparent-citizen: Oh, absolutely. Fixed in https://github.com/edmundhung/conform/pull/1188/changes/6987b5912083b0d0c0bb9ed50bfac8046a965e65.

packages/conform-dom/util.ts
transparent-citizen: Agreed, I put this in https://github.com/edmundhung/conform/pull/1188/changes/6987b5912083b0d0c0bb9ed50bfac8046a965e65.

packages/conform-dom/util.ts
transparent-citizen: What I suggested above is an oversimplification. The general case of getting rid of the `i` flag is non-trivial, since it would have to support Unicode case folding rules.

For instance, consider a Greek sigma (Σ) with its two lowercase forms σ and ς that follow positional rules, the latter form being used at the end of a word. To rewrite `/Σ/i` without `i`, we'd have to produce `/[Σσς]/`.

Another issue is that some mappings are asymmetric:
```js
/s/ui.test("ſ") // ✅
/[sS]/u.test("ſ") // ❌
```

Currently, I think we're better off not handling case-insensitive regexes at all.

We could make them best-effort cases and silently strip the flag, but this would change semantics and lead to misbehaving programs. We could also explicitly map them to `undefined`. This fails fast and gives the user a chance to rewrite the regex without `i`.

Do you have a preference how we should handle this?

Arguably, a better solution would be to get something like an `insensitive` attribute into the HTML spec for `<input>`.

packages/conform-dom/tests/util.test.ts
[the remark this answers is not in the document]
transparent-citizen: By stripping all flags from case-sensitive regexes in dfff9e0894f78773a5da00632ce1c05e40e3d6f5, `v` is not a special case anymore.

packages/conform-dom/util.ts
edmundhung: > Is it at all possible to enter the catch case? Given that the inputs to serializeHtmlPattern are RegExp objects with valid patterns, and we're placing them in isolated slots using (?:...), it couldn't possibly yield an unparseable regex, can it?

One case GPT 5.4 suggests is that a regex like `/[()]/` would become invalid under `v` as some characters need to be escaped. It's gonna be hard to escape on behalf of the users as `/\[\(\)\]/v` would change the meaning from either `(` or `)` to a literal string `[()]`. Manually valdiating the regex with a v flag seems like a safe fallback anyway.

> Do you have a preference how we should handle this?

Fails fast makes sense to me 👍🏼

transparent-citizen: Good catch! I've added the protection against invalid patterns in 21f66d8.

fc5a6da3c125531614338de75c014347f11b9fee contains the changes to make `i` fail fast and support all other flags.

transparent-citizen:
This should be good to go, feel free to squash when you're ready. Thanks for keeping the feedback loop so short! 💯

packages/conform-dom/util.ts
```diff
@@ -113,3 +113,34 @@ export function getTypeName(value: unknown): string {
...
+
+	const serializedPattern = `^${patterns.map((p) => `(?=.*(?:${p.source}))`).join('')}.*$`;
+
+	try {
+		new RegExp(serializedPattern, 'v');
+	} catch {
+		return undefined;
+	}
```
edmundhung, line 143: This seems to suggest we are going for all or nothing:

should we filter out only invalid regex and keep serializing those valid one?

packages/conform-dom/util.ts
```diff
@@ -113,3 +113,34 @@ export function getTypeName(value: unknown): string {
...
+export function serializeHtmlPattern(
+	pattern: RegExp | RegExp[],
+): string | undefined {
+	const patterns = Array.isArray(pattern) ? pattern : [pattern];
+
+	if (patterns.length === 0 || patterns.some((p) => p.flags.includes('i'))) {
+		return undefined;
+	}
+
+	const serializedPattern = `^${patterns.map((p) => `(?=.*(?:${p.source}))`).join('')}.*$`;
```
edmundhung, line 137: I just notice this might impact backreference as we are concatenating the source. 😅

I'm not sure it's worth adding more heuristic checks here just to catch edge cases like this. Maybe documenting this would be enough?

packages/conform-dom/util.ts
transparent-citizen: Let's consider a schema with two patterns, for instance:

| RegExp | Description |
|----|----|
| `/[a-z]+/i` | any Latin letter (case-insensitive) |
| `/[0-9]/` | any digit  |

Since we can't serialize `i`, stripping it down to

| RegExp | Description |
|----|----|
| `/[0-9]/` | any digit  |

would change the semantics: The HTML pattern would accept `7`, but reject `a`. This is a problem:

1. The behavior is unexpected and does not reflect user intent. It creates a mismatch between what's written in the user's code and what actually happens.
2. JS-based schema validation will happily use both regexes and therefore produce a validity mismatch for the same input, which adds to the confusion.

It's a little disappointing that there's currently no right way to make `i` work. But not creating a pattern when Conform sees that it can't is less of a hindrance to the user than it deliberately misinterpreting the user's request and coming up with a pattern we know they didn't ask for.

packages/conform-dom/util.ts
transparent-citizen: Whoops 😅 Great catch. Numbered backreferences will break for sure, but named backreferences (`\k<name>`) can be used as a workaround.

The only limitation would be that group names would have to be made unique among input regexes.

If we parsed regexes to ASTs, we could work around either limitation and give each group a unique prefix and also shift numbered backreferences.

I've documented it in 2ee9578, but I wasn't sure where. The docs for constraints are currently split by validation library, so I had to duplicate it. If you prefer moving it somewhere else and/or updating the text, feel free.

packages/conform-dom/util.ts
edmundhung: Ah, right. That makes sense. 👍🏼

packages/conform-dom/util.ts
edmundhung: I have slightly updated the docs. But otherwise they looks great 👍🏼

edmundhung (approved):
Great feature! LGTM

transparent-citizen:
I just saw Conform put this masterpiece onto an `<input type="datetime-local">` via `z.iso.datetime({ local: true })` and was pleasantly surprised to see it parse correctly, thought you'd appreciate it as well. 😄

Thanks for merging!

edmundhung:
Nice! Good to see that works out very well 🔥

PR #1211, opened 2026-05-01
fix(conform-zod): Zod 4.4 coercion optionality handling

### LINKED ISSUE #1210: Checkbox bug with zod v4.4

Previously, a zod definiton such as this one
        checkboxName: z.boolean().default(false)
was working totally fine for a checkbox, but after the upgrade to zod v4.4, when trying to submit with the checkbox unchecked - the following error is produced: 'Invalid input: expected nonoptional, received undefined'

I've tried the following, with the error, curiously, staying the same:
z.boolean().optional()
z.boolean().default(false).optional()
z.boolean().optional().default(false)
z.boolean().optional().default(false).optional()

Clearly the zod upgrade somehow messed up the validation.

### Conform version

v1.19.1

https://codesandbox.io/p/devbox/hn2hjj

Chrome

### Screenshots or Videos

_No response_

### Additional context

_No response_

chimame:
## Overview

Fix #1210

Fixes Zod v4 coercion so schemas wrapped with preprocessing pipes preserve their input/output optionality metadata.

Zod 4.4 tightened object property handling around missing keys. Because Conform wraps schemas with `pipe(transform(...), target)` during coercion, optional/default metadata could be lost on the outer schema, causing missing optional/defaulted object fields to be reported as nonoptional.

This change copies `_zod.optin` and `_zod.optout` from the wrapped target schema to the generated pipe schema, and preserves the existing behavior for required arrays/objects that materialize missing values as `[]` or `{}`.

<details>

This PR preserves Zod v4.4 input/output optionality metadata when Conform wraps schemas for coercion, preventing optional/defaulted object and array fields from being reported as required.

Key changes:
- Introduces pipeWithOptionality() to compose pipe(transform(...), target) while copying target._zod.optin/_zod.optout onto the pipe schema.
- Refactors coerceType() to use a CoerceTypeOptions object and apply pipeWithOptionality across coercion paths.
- Adds coerceObjectShapeEntry() to mark per-property coerced schemas optional when the underlying type can materialize missing arrays/objects, ensuring omitted fields materialize as [] or {}.
- Applies optionality handling to discriminated-union variants and avoids leaking optionality metadata to reused schema instances.
- Adds tests for discriminated unions and object reuse scenarios and includes a small type fix and a changeset for a patch release of @conform-to/zod.

</details>

### DIFF
--- a/packages/conform-zod/tsconfig.json
+++ b/packages/conform-zod/tsconfig.json
@@ -4,6 +4,11 @@
 		"target": "ES2022",
 		"module": "ES2022",
 		"moduleResolution": "Bundler",
+		"baseUrl": ".",
+		"paths": {
+			"zod/v4": ["./node_modules/zod-v4/v4"],
+			"zod/v4/*": ["./node_modules/zod-v4/v4/*"]
+		},
 		"allowSyntheticDefaultImports": true,
 		"noUncheckedIndexedAccess": true,
 		"strict": true,
--- a/packages/conform-zod/tsconfig.json
+++ b/packages/conform-zod/tsconfig.json
@@ -6,6 +6,10 @@
 		"moduleResolution": "Bundler",
 		"baseUrl": ".",
 		"paths": {
+			// Keep v4 sources and tests on the zod-v4 alias. The zod devDependency is
+			// still v3 for v3 support, so resolving zod/v4 normally would use the v4
+			// compatibility types bundled with zod@3. This alias can be removed when
+			// v3 support is dropped and zod itself is upgraded to v4.
 			"zod/v4": ["./node_modules/zod-v4/v4"],
 			"zod/v4/*": ["./node_modules/zod-v4/v4/*"]
 		},
--- a/packages/conform-zod/v4/coercion.ts
+++ b/packages/conform-zod/v4/coercion.ts
@@ -248,9 +248,7 @@ function materializesMissingValue(type: $ZodType): boolean {
 			return materializesMissingValue(pipeDef.out);
 		}
 
-		if (pipeDef.out._zod.def.type === 'transform') {
-			return materializesMissingValue(pipeDef.in);
-		}
+		return materializesMissingValue(pipeDef.in);
 	}
 
 	return def.type === 'array' || def.type === 'object';
--- a/packages/conform-zod/v4/coercion.ts
+++ b/packages/conform-zod/v4/coercion.ts
@@ -230,6 +230,23 @@ function materializesMissingValue(type: $ZodType): boolean {
 	return def.type === 'array' || def.type === 'object';
 }
 
+function withOptionalInput(type: $ZodType): $ZodType {
+	const schema = Object.create(Object.getPrototypeOf(type)) as $ZodType;
+	const descriptors = Object.getOwnPropertyDescriptors(type);
+
+	delete descriptors._zod;
+	Object.defineProperties(schema, descriptors);
+	Object.defineProperty(schema, '_zod', {
+		value: Object.create(type._zod, {
+			optin: {
+				value: 'optional',
+			},
+		}),
+	});
+
+	return schema;
+}
+
 function coerceObjectShapeEntry(
 	type: $ZodType,
 	options: CoerceTypeOptions,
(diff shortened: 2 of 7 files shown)


packages/conform-zod/v4/coercion.ts
```diff
@@ -191,29 +208,74 @@ function typeCheckTransform(defType: string): $ZodType {
...
+	while (
+		def.type === 'optional' ||
+		def.type === 'nonoptional' ||
+		def.type === 'default' ||
+		def.type === 'prefault' ||
+		def.type === 'catch' ||
+		def.type === 'nullable'
+	) {
+		def = (def as unknown as { innerType: $ZodType }).innerType._zod.def;
+	}
```
edmundhung: I am actually still a bit lost with the logic. But it seems like you need to also handle `pipe` to support `z.preprocess` or `.transform()` 🤔 

```suggestion
	if (
		def.type === 'optional' ||
		def.type === 'nonoptional' ||
		def.type === 'default' ||
		def.type === 'prefault' ||
		def.type === 'catch' ||
		def.type === 'nullable'
	) {
		return materializesMissingValue(def as unknown as { innerType: $ZodType }).innerType);
	}
	
	if (def.type === 'pipe') {
		if (def.in._zod.def.type === 'transform') {
			return materializesMissingValue(def.out);
		}
		if (def.out._zod.def.type === 'transform') {
			return materializesMissingValue(def.in);
		}
	}
```

chimame: Good point, I agree this should handle transform-only pipes as well.

`materializesMissingValue()` is only trying to detect whether a missing object field can be materialized by Conform’s coercion into `[]` or `{}`. For `z.preprocess(fn, z.array(...))`, Zod represents that as `pipe(transform, array)`, and for `z.array(...).transform(fn)` it is `pipe(array, transform)`, so the current `innerType` unwrap misses those cases.

I’ll update this to treat a pipe as transparent only when one side is a `transform`, then inspect the other side. I’ll also adjust the suggested code slightly for the Zod core types and add regression coverage for both preprocess and transform cases.

packages/conform-zod/v4/coercion.ts
```diff
@@ -191,29 +208,88 @@ function typeCheckTransform(defType: string): $ZodType {
...
+		const pipeDef = def as unknown as { in: $ZodType; out: $ZodType };
+
+		if (pipeDef.in._zod.def.type === 'transform') {
+			return materializesMissingValue(pipeDef.out);
+		}
+
+		if (pipeDef.out._zod.def.type === 'transform') {
+			return materializesMissingValue(pipeDef.in);
+		}
+	}
```
edmundhung, line 252: GPT 5.5 suggests this to fix raw `.pipe()` support. I can't really think of a realworld usecase for `.pipe()` but I guess we do want to support all kind of schema anyway 😅 

```suggestion
	if (def.type === 'pipe') {
		const pipeDef = def as unknown as { in: $ZodType; out: $ZodType };

		if (pipeDef.in._zod.def.type === 'transform') {
			return materializesMissingValue(pipeDef.out);
		}

		return materializesMissingValue(pipeDef.in);
	}
```

edmundhung (approved):
Thanks for working on this! Should we also update the `"zod-v4"` version in the package to latest so we are sure it works with that? It does mean we are no longer verifying this works with older zod 4 version. But most people care about latest zod version support when they update a package.

packages/conform-zod/tsconfig.json
```diff
@@ -4,6 +4,15 @@
...
 		"module": "ES2022",
 		"moduleResolution": "Bundler",
+		"baseUrl": ".",
+		"paths": {
+			// Keep v4 sources and tests on the zod-v4 alias. The zod devDependency is
+			// still v3 for v3 support, so resolving zod/v4 normally would use the v4
+			// compatibility types bundled with zod@3. This alias can be removed when
+			// v3 support is dropped and zod itself is upgraded to v4.
+			"zod/v4": ["./node_modules/zod-v4/v4"],
+			"zod/v4/*": ["./node_modules/zod-v4/v4/*"]
```
edmundhung, line 14: Good catch 👍🏼

edmundhung (approved):
Let's release this fix :)

Issue #1216, opened 2026-05-15
coerceFormValue returns any in 1.19.2 — dist/v4/future.d.ts incorrectly references .cjs module paths


leonardo-ornelas:
### Describe the bug and the expected behavior

After upgrading @conform-to/zod from 1.19.0 to 1.19.2, coerceFormValue (imported from @conform-to/zod/v4/future) returns any instead of the inferred schema type.

Root cause
dist/v4/future.d.ts (the ESM type entry, used when "types" in the package exports resolves to .d.ts) changed its internal type-level imports to reference CJS module specifiers:

1.19.0 (dist/v4/future.d.ts):

```
export declare const coerceFormValue: <Schema extends import("zod/v4/core").$ZodType>(type: Schema) =>
  import("zod/v4").ZodType<import("zod/v4").infer<Schema>, import("zod/v4").input<Schema>>;

```
1.19.2 (dist/v4/future.d.ts):

```
export declare const coerceFormValue: <Schema extends import("zod/v4/core/schemas.cjs").$ZodType>(type: Schema) =>
  import("zod/v4/index.cjs").ZodType<import("zod/v4/index.cjs").infer<Schema>, import("zod/v4/index.cjs").input<Schema>>;
```

In an ESM project, z from zod resolves through zod/v4/core (→ schemas.d.ts), while the constraint now checks against zod/v4/core/schemas.cjs (→ schemas.d.cts). TypeScript treats these as distinct module identities, the generic constraint Schema extends $ZodType fails to narrow, and the return type collapses to any.

### Conform version

v1.19.2

### Steps to Reproduce the Bug or Issue

```
import { coerceFormValue } from '@conform-to/zod/v4/future'
import { z } from 'zod'

const schema = coerceFormValue(z.object({ age: z.number() }))
//    ^? ZodType<any, any>  ← should be ZodType<{ age: number }, { age: string }>
```

### Screenshots or Videos

Exxample:

<img width="793" height="201" alt="Image" src="https://github.com/user-attachments/assets/1a46771a-6445-4140-8735-01e373206fbb" />

edmundhung:
It might be related to the tsconfig change in #1211. I will take a look later today and try to a get a patch out over the weekend.

edmundhung:
Thanks for the report. This should be fixed in [1.19.3](https://github.com/edmundhung/conform/releases/tag/v1.19.3).

PR #1217, opened 2026-05-15
fix(conform-zod): avoid cjs paths in zod v4 declarations

chimame:
## Overview

#1216 

<details>

This patch prevents Zod v4 path aliases from leaking into the built TypeScript declarations for the `@conform-to/zod` package by clearing compilerOptions.paths in the package build tsconfig. The package-level tsconfig still keeps the "zod/v4" path alias for editor/test resolution; the build config explicitly resets paths so declaration emission doesn't use the aliased (CJS) import paths.

Example of the added build config:

```json
{
  "extends": "./tsconfig",
  "compilerOptions": {
    "outDir": "./dist",
    "emitDeclarationOnly": true,
    "noEmit": false,
    "paths": {
      // Do not inherit zod-v4 path aliases.
    }
  },
  "exclude": ["**/tests", "dist"]
}
```

</details>

[Review Change Stack](https://app.coderabbit.ai/change-stack/edmundhung/conform/pull/1217?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/conform-zod/tsconfig.build.json
+++ b/packages/conform-zod/tsconfig.build.json
@@ -4,7 +4,9 @@
 		"outDir": "./dist",
 		"emitDeclarationOnly": true,
 		"noEmit": false,
-		"paths": {}
+		"paths": {
+			// Do not inherit zod-v4 path aliases.
+		}
 	},
 	"exclude": ["**/tests", "dist"]
 }
--- a/packages/conform-zod/tsconfig.build.json
+++ b/packages/conform-zod/tsconfig.build.json
@@ -3,7 +3,8 @@
 	"compilerOptions": {
 		"outDir": "./dist",
 		"emitDeclarationOnly": true,
-		"noEmit": false
+		"noEmit": false,
+		"paths": {}
 	},
 	"exclude": ["**/tests", "dist"]
 }
--- a/.changeset/selfish-wasps-confess.md
+++ b/.changeset/selfish-wasps-confess.md
@@ -1,0 +1,5 @@
+---
+'@conform-to/zod': patch
+---
+
+fix(conform-zod): avoid cjs paths in zod v4 declarations


edmundhung:
Thanks for the quick fix 🫶 

Shall we add a `tsconfig.json` with the custom `paths` inside `tests`? I think it will be more intuitive.

chimame:
I tried moving the custom paths into `packages/conform-zod/v4/tests/tsconfig.json`, but VS Code still reports type errors because the test files and the imported v4 source files end up resolving zod/v4 from different Zod installations. The tests resolve zod/v4 through zod@4, while the source files fall back to the zod/v4 compatibility types from zod@3.

So I think keeping the paths in the package tsconfig is intentional for the dev/editor type environment. The declaration issue is specific to build output, and overriding paths in tsconfig.build.json fixes that without breaking editor behavior.

```json
// packages/conform-zod/v4/tests/tsconfig.json
{
  "extends": "../../tsconfig.json",
  "compilerOptions": {
    "baseUrl": "../../",
    "paths": {
      "zod/v4": ["./node_modules/zod-v4/v4"],
      "zod/v4/*": ["./node_modules/zod-v4/v4/*"]
    }
  },
  "include": ["../**/tests/**/*.ts", "../**/tests/**/*.tsx"]
}
```

```json
// packages/conform-zod/tsconfig.json
{
	"compilerOptions": {
		"lib": ["ES2022", "DOM", "DOM.Iterable"],
		"target": "ES2022",
		"module": "ES2022",
		"allowSyntheticDefaultImports": true,
		"noUncheckedIndexedAccess": true,
		"strict": true,
		"declaration": true,
		"declarationMap": true,
		"composite": true,
		"noEmit": true,
		"skipLibCheck": true
	},
	"exclude": ["v4/tests", "dist"]
}
```

image

packages/conform-zod/tsconfig.build.json
```diff
@@ -3,7 +3,8 @@
 	"compilerOptions": {
 		"outDir": "./dist",
 		"emitDeclarationOnly": true,
-		"noEmit": false
+		"noEmit": false,
+		"paths": {}
```
edmundhung: Nit: maybe worth including a comment here

```suggestion
		"paths": {
		  // Do not inherit zod-v4 path aliases.
		}
```

edmundhung (approved):
That makes sense. Let's go with this :)

transparent-citizen:
Looks like there are no type tests for `coerceFormValue` yet. Maybe now is a good opportunity? Something like this could prevent this from breaking again:

```ts
test('preserves string schema types', () => {
	const schema = z.string();
	const coerced = coerceFormValue(schema);

	expectTypeOf<z.output<typeof coerced>>().toEqualTypeOf<
		z.output<typeof schema>
	>();
	expectTypeOf<z.input<typeof coerced>>().toEqualTypeOf<
		z.input<typeof schema>
	>();
});
```

edmundhung:
> Looks like there are no type tests for `coerceFormValue` yet. Maybe now is a good opportunity? Something like this could prevent this from breaking again:

I can add that after the PR as well. @chimame feel free to merge anytime. 👍🏼

chimame:
I agree type coverage for coerceFormValue would be useful, but I don’t think a source-level type test would catch this particular regression. The implementation types were fine before build; the issue was introduced by declaration emit, where `dist/v4/future.d.ts` ended up referencing `zod/v4/index.cjs` and `zod/v4/core/schemas.cjs`.

A regular expectTypeOf test against the source would still pass because it runs under our dev tsconfig paths. To prevent this regression, I think we would need either a post-build check against `dist/v4/future.d.ts`, or a consumer-style type test that imports from the built package output.

PR #1222, opened 2026-05-17
chore: optimize vitest config

edmundhung:
<details>

Refactors Vitest configurations across four package subdirectories to inline `projects` definitions directly within `defineConfig()` rather than exporting them as a separate named constant. This consolidation simplifies the config structure and eliminates the separate `defineProject()` calls.

**Key changes:**
- Removes `export const projects = [...]` declarations from all four config files
- Projects array is now defined inline within `test: { projects: [...] }`
- Removes custom `browser.api` port assignments that were previously specified
- In `conform-react`, consolidates `optimizeDeps` (React JSX dev runtime prebundle) to the top-level config instead of per-project

All test behavior and coverage patterns remain unchanged—only the structural organization of the configuration is refined.

</details>

[Review Change Stack](https://app.coderabbit.ai/change-stack/edmundhung/conform/pull/1222?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/conform-dom/vitest.config.ts
+++ b/packages/conform-dom/vitest.config.ts
@@ -1,55 +1,54 @@
 import { fileURLToPath } from 'node:url';
-import { defineConfig, defineProject } from 'vitest/config';
+import { defineConfig } from 'vitest/config';
 import { playwright } from '@vitest/browser-playwright';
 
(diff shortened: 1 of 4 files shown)


edmundhung:
@coderabbitai review

PR #1233, opened 2026-06-09
Preserve empty string array values in submissions

### LINKED ISSUE #839: `parseWithZod` fails to parse Arrays containing empty strings (`''`) correctly

When an array in formData contains an empty string (`''`) as the first element, it gets trimmed, and the resulting payload has fewer elements than expected. As a result, the indices in parseResult.value do not match the original formData.
This behavior also affects validation errors when using useForm, making it difficult to pinpoint which input has an error.

### Conform version

v1.1.5

```typescript
test("parseWithZod with array", () => {
  const formData = new FormData();

  formData.append("animals", "");
  formData.append("animals", "dog");
  formData.append("animals", "");
  formData.append("animals", "cat");

  const result = parseWithZod(formData, {
    schema: z.object({
      animals: z.array(z.string().optional()),
    }),
  });

  expect(result.status).toBe("success");
  expect(result.payload).toEqual({
    animals: ["", "dog", "", "cat"],
  });
});
```

```
AssertionError: expected { animals: [ 'dog', '', 'cat' ] } to deeply equal { animals: [ '', 'dog', '', 'cat' ] }

- Expected
+ Received

  Object {
    "animals": Array [
-     "",
      "dog",
      "",
      "cat",
    ],
  }
```

Chrome

### Screenshots or Videos

_No response_

### Additional context

This issue occurs because `''` is treated as falsy. As a result, setting values into the context is skipped until a concrete value is provided.

https://github.com/edmundhung/conform/blob/b0fb91dbab474c7a150d5f56cb71dbb92888779c/packages/conform-dom/submission.ts#L103-L104

morgan-coded:
`parseWithZod` could drop an empty string inside an array field because that value was treated like it was absent. That can lose or shift array entries in the parsed submission payload.

This narrows the presence check so only `undefined` is treated as missing. I checked it with the focused `conform-dom` submission test, changed-file `oxlint`, changed-file `oxfmt --check`, and `git diff --check`.

Closes #839

<details>

## Changes

Modified the presence check in `packages/conform-dom/submission.ts` to preserve empty string values in array fields. The reducer now treats only `undefined` as missing (previously treated any falsy value as missing):

```typescript
// Before: !prev (treats empty string as missing)
// After: prev === undefined (only undefined is missing)
```

This ensures empty string elements in arrays are not dropped during form submission parsing, maintaining correct array indices and element ordering.

## Additional updates

- Added Changesets entry documenting the behavior change
- Added test verifying empty strings are preserved within array fields during parsing

</details>

### DIFF
--- a/.changeset/fix-array-empty-string.md
+++ b/.changeset/fix-array-empty-string.md
@@ -1,0 +1,5 @@
+---
+'@conform-to/dom': patch
+---
+
+Preserve empty string values inside array fields when building submission payloads.
--- a/packages/conform-dom/submission.ts
+++ b/packages/conform-dom/submission.ts
@@ -101,7 +101,7 @@ export function getSubmissionContext(
 			context.payload,
 			name,
 			(prev: unknown) => {
-				if (!prev) {
+				if (prev === undefined) {
 					return next;
 				} else if (Array.isArray(prev)) {
 					return prev.concat(next);
--- a/packages/conform-dom/tests/submission.test.ts
+++ b/packages/conform-dom/tests/submission.test.ts
@@ -281,3 +281,29 @@ test('parse()', () => {
 		},
 	});
 });
+
+test('parse() preserves empty strings in array fields', () => {
+	const submission = parse(
+		createFormData([
+			['animals', ''],
+			['animals', 'dog'],
+			['animals', ''],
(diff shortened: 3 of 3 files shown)


edmundhung (approved):
Thanks!

PR #1237, opened 2026-06-15
Preserve lastResult errors without initialValue

### LINKED ISSUE #1003: Missing error messages when `initialValue` is undefined in the last result object

I was mocking the result of a form submission in a test but couldn’t get the expected render. After debugging, I realized that adding an empty object for `initialValue` in the mocked object fixes the issue. Note that the `initialValue` in `SubmissionResult` is optional.

Here’s the mock I created for the React Router/Remix component:

```typescript
    const Stub = createRoutesStub([
      {
        path: '/',
        Component: ComponentUnderTheTest,
        action(): SubmissionResult<string[]> {
          return {
            status: 'error',

            // it works with an empty object, but if I comment this line it will cause the test to fail
            initialValue: {},

            error: {
              name: ['some error'],
            },
            fields: ['name'],
          };
        },
      },
    ]);

   expect(screen.getByRole('textbox', {
        description: 'some error',
      })).toBeInTheDocument();
```

### Conform version

v1.8.2

Use the following snippet in a React component:

```typescript

  const lastResult = useMemo(() => {
    return {
      status: 'error',

      // uncoment this line for expected behavior 
      // initialValue: {},
      error: {
        name: ['some error message'],
      },
      fields: ['name'],
    }
  }, []);

    lastResult,
  });

  console.log('Error message', fields.name.errors);
```

It’s expected to print `some error message` in the logs, but it doesn’t.

Chrome

### Screenshots or Videos

_No response_

### Additional context

_No response_

morgan-coded:
`lastResult` errors could be missed when `initialValue` was omitted.
This keeps `null` as the explicit reset case, while letting `undefined` continue through the existing initial-state path so returned errors are still picked up. I checked it with the focused repro, the `@conform-to/dom` node tests passing 49/49, typecheck, oxlint, oxfmt, and `git diff --check`.
Fixes #1003

<details>

## Changes

- **Core fix** (`packages/conform-dom/form.ts`): Updated `report` reset handling to treat **`initialValue: null`** as the explicit reset signal, while **`initialValue: undefined`** (omitted) follows the existing initial-state path—so **`lastResult` errors aren’t dropped**.
  ```ts
  // conceptual change
  if (result.initialValue === null) reset();
  // otherwise preserve prior lastResult error path behavior
  ```
- **Tests** (`packages/conform-dom/tests/form.node.test.ts`): Added node/DOM-focused coverage to verify error preservation when `initialValue` is omitted, correct `submissionStatus`/`error` updates from `onUpdate` `lastResult`, and reset/clearing when `lastResult.initialValue` is `null`.
- **Release note** (`.changeset/quiet-errors-flow.md`): Patch Changesets entry documenting that `lastResult` errors are preserved when `initialValue` is omitted.

</details>

### DIFF
--- a/.changeset/quiet-errors-flow.md
+++ b/.changeset/quiet-errors-flow.md
@@ -1,0 +1,5 @@
+---
+'@conform-to/dom': patch
+---
+
+Preserve `lastResult` errors when `initialValue` is omitted.
--- a/packages/conform-dom/form.ts
+++ b/packages/conform-dom/form.ts
@@ -944,7 +944,10 @@ export function createFormContext<
 			return;
 		}
 
-		const initialValue = result.initialValue ?? meta.initialValue;
+		const initialValue =
+			typeof result.initialValue !== 'undefined'
+				? result.initialValue
+				: meta.initialValue;
 		const error = Object.entries(result.error ?? {}).reduce<
 			Record<string, FormError>
 		>((result, [name, newError]) => {
@@ -966,7 +969,8 @@ export function createFormContext<
 			pendingIntents,
 			isValueUpdated: false,
 			submissionStatus: result.status,
-			value: initialValue,
+			value:
+				typeof result.initialValue !== 'undefined' ? initialValue : meta.value,
 			validated: {
 				...meta.validated,
 				...result.state?.validated,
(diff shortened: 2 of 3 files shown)


edmundhung (approved):
Sorry for the late review! I will get this release now :)

PR #1239, opened 2026-06-17
fix(zod): support z.readonly() in coerceFormValue and getZodConstraint

chatman-media:
## Problem

Fields wrapped with [`z.readonly()`](https://zod.dev/api?id=readonly) are not handled by the Zod adapter. The `readonly` wrapper is treated as an opaque/unknown type, so:

1. **`coerceFormValue` skips coercion** for the inner type. A form value for `z.number().readonly()` arrives as a string and validation fails:

   ```ts
   const schema = coerceFormValue(z.object({ age: z.number().readonly() }));
   schema.safeParse({ age: '42' });
   // ✗ { code: 'invalid_type', expected: 'number', received: 'string', path: ['age'] }
   ```

   The same happens for a `z.object({ ... }).readonly()` field (the empty-object default and nested coercion are skipped).

2. **`getZodConstraint` drops the field's validation attributes**. A `z.string().min(5).readonly()` field loses its `minLength`:

   ```ts
   getZodConstraint(z.object({ name: z.string().min(5).readonly() }));
   // before: { name: { required: true } }   ← minLength missing
   ```

## Fix

`readonly` is a transparent wrapper, just like `nullable`, `catch`, and `brand`, which the adapter already unwraps. This PR unwraps it the same way in both `coerceFormValue` (rebuilding the wrapper around the coerced inner type) and `getZodConstraint` (recursing into `innerType`).

In the constraint walker the `readonly` branch is placed alongside the other transparent wrappers that run *before* the root-level guard, so a schema whose root is `z.object({ ... }).readonly()` is supported too.

Applied consistently across all three adapters: `default`, `/v3`, and `/v4` (both the chainable `.readonly()` and the `zod-mini` `readonly(...)` form).

## Tests

- New `coercion/schema/readonly.test.ts` for `default`, `v3`, and `v4` (mirrors the existing `brand.test.ts`), covering every primitive, nested objects, optional combinations, and (for v4) the `zod-mini` form.
- New `readonly` case in each `constraint.test.ts`, covering constraint derivation through readonly fields including a readonly root object and a readonly + optional combination.

All tests fail before the change and pass after. The full `@conform-to/zod` node test suite passes (`208 passed | 2 skipped`) with no type errors.

A changeset is included (`@conform-to/zod`, patch).

<details>

This PR teaches the Conform Zod adapters to treat `z.readonly()` as a transparent wrapper rather than an opaque schema.

- **Coercion:** adapters now detect `ZodReadonly`, recursively coerce its `innerType`, and then re-wrap with `readonly`.
- **Constraints:** `getZodConstraint` now derives constraint attributes from the wrapped `innerType` (so they aren’t dropped).
- **v4 missing-value behavior:** readonly participates in the existing “materialize missing values” logic (important for empty-object/empty-array coercion paths).
- **Adapter coverage:** implemented in **default**, **/v3**, and **/v4** (including `zod-mini` readonly forms).

Example of preserved constraints:
```ts
z.string().min(5).readonly() // keeps minLength constraint in generated constraints
```

Tests added across default/v3/v4 validate readonly coercion and constraint extraction for primitives, nested objects, readonly root objects, readonly/optional combinations, and constraint preservation (including `File`-instance handling where applicable). A changeset updates `@conform-to/zod`.

</details>

### DIFF
--- a/.changeset/coerce-readonly-schema.md
+++ b/.changeset/coerce-readonly-schema.md
@@ -1,0 +1,7 @@
+---
+'@conform-to/zod': patch
+---
+
+Support `z.readonly()` schemas in `coerceFormValue` and `getZodConstraint`.
+
+Previously, the `readonly` wrapper was treated as an opaque type, so coercion was skipped for any field defined with `.readonly()` (e.g. `z.number().readonly()` would fail with `expected number, received string`) and `getZodConstraint` dropped the field's validation attributes (such as `minLength`). The wrapper is now unwrapped like other transparent wrappers (`.nullable()`, `.brand()`), so its inner type is coerced and its constraints are derived correctly. This applies to the default, `/v3`, and `/v4` adapters.
--- a/packages/conform-zod/default/coercion.ts
+++ b/packages/conform-zod/default/coercion.ts
@@ -13,7 +13,6 @@ import {
 	any,
 	ZodCatch,
 	ZodBranded,
-	ZodReadonly,
 	ZodDefault,
 } from 'zod';
 import type {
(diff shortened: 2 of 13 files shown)


edmundhung (approved):
Sorry for the late review. I have slightly adjusted the implementation for the default entrypoint due to zod version compatibility issue. But otherwise they look great to me.

Thanks! 🙏🏼

PR #1246, opened 2026-07-19
feat: align future APIs on targetValue

edmundhung:
## Summary
- rename the resolveSubmission result from value to targetValue
- add targetValue to report while retaining value as a deprecated alias for one minor release
- make targetValue take precedence when both report options are provided
- update runtime call sites, tests, examples, docs, and migration-focused changeset notes

resolveSubmission is an immediate breaking change in the experimental future API. report callers have a one-minor migration window before value is removed.

## Testing
- pnpm --filter @conform-to/dom build
- pnpm --filter @conform-to/dom test --run tests/formdata.test.ts
- pnpm --filter @conform-to/dom typecheck
- pnpm --filter @conform-to/react test --run tests/resolveSubmission.browser.test.tsx
- pnpm --filter @conform-to/react typecheck
- pnpm --filter react-router-example typecheck
- targeted oxlint and oxfmt checks

Astro check still reports two pre-existing nullability errors in examples/astro/src/features/login.tsx.

<details>

Renames the experimental future `report()` option from `value` to `targetValue` and updates `resolveSubmission()` to return `{ intent, targetValue }` instead of `{ intent, value }`. `value` remains as a deprecated alias for one minor release, and `targetValue` takes precedence when both are provided; `hideFields` behavior is documented in terms of the target payload. Updates include runtime plumbing, React/DOM call sites, docs, and tests (runtime + type tests asserting `value` is rejected via `@ts-expect-error`).

```ts
const { intent, targetValue } = await resolveSubmission(...);

return report(submission, {
  intent,
  targetValue,
});
```

Also adds changeset notes to publish the rename and temporary alias guidance (with example destructuring updates).
</details>

### DIFF
--- a/.changeset/clean-dodos-target.md
+++ b/.changeset/clean-dodos-target.md
@@ -3,4 +3,22 @@
 '@conform-to/react': minor
 ---
 
-Rename the future `report()` option and `resolveSubmission()` result from `value` to `targetValue` for consistency with `SubmissionResult.targetValue`.
+Align the experimental future APIs on `targetValue` for the form value produced by resolving an intent and applied by a submission result.
+
+The `resolveSubmission()` result has a breaking change:
+
+```diff
+-const { intent, value } = resolveSubmission(submission);
++const { intent, targetValue } = resolveSubmission(submission);
+```
+
+The `report()` option is renamed too, but `value` remains available as a deprecated alias for one minor release:
+
+```diff
+ report(submission, {
+-  value,
++  targetValue,
+ });
+```
+
+`value` will be removed from `report()` in the next minor release. If both `value` and `targetValue` are provided during migration, `targetValue` takes precedence.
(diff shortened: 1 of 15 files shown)


transparent-citizen:
The bikeshedding over this already bikeshedded report option will never end, and I think that's a good thing. 😄

What's the relation between `submission.payload` and `resolveSubmission(submission).targetValue`? Which should consumers parse when reading submitted values?

edmundhung:
`submission.payload` is the submitted data. The `resolveSubmission` API is only needed if you want the server to process the intent for progressive enhancement.

It applies the intent to `submission.payload` and returns a `targetValue`. e.g. If you call `intent.update({ name: 'foo', value: '' })`, then `targetValue` would be `submission.payload` with `foo` set to an empty string.

In this case, you should validate `targetValue` and pass it to `report`. Otherwise, you can continue to validate `submission.payload` directly.

PR #1258, opened 2026-07-24
fix(conform-react): fix faulty case when inferring schema output type

transparent-citizen:
An unexpected schema still produces an output, it's just that we don't know its type. `undefined` is the wrong fallback type. It makes more sense for `InferOutput` to fallback to `unknown`.

<details>

- Updated `packages/conform-react/future/types.ts` to make `InferOutput<Schema>` return `undefined` when the schema type is actually `undefined`, while changing the unsupported/custom-schema fallback from `undefined` to `unknown`.
- Adjusted the `FormOptions` default for the `Schema` generic from `unknown` to `undefined` to align the “schema not provided” case with the new `InferOutput` behavior.
- Updated `packages/conform-react/tests/types.test-d.ts` type tests to validate `onValidate`’s `schemaValue` inference (`undefined` for the defaulted `FormOptions`, `unknown` for the configured variant).
- Added a patch Changeset for `@conform-to/react` documenting the type inference fix for unexpected/unsupported custom schemas.

```ts
// Illustration of the intended behavior
type A = InferOutput<undefined>; // undefined
type B = InferOutput<SomeUnsupportedSchema>; // unknown
```

</details>

### DIFF
--- a/.changeset/fuzzy-schemas-report.md
+++ b/.changeset/fuzzy-schemas-report.md
@@ -1,0 +1,5 @@
+---
+'@conform-to/react': patch
+---
+
+Fix output inference for unrecognized schemas to fall back to `unknown` while preserving `undefined` when no schema is provided.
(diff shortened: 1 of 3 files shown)


edmundhung (approved):
It was a trick I did so useForm fallback to undefined when no schema is provided. But I guess it's time to fix it properly 😄

transparent-citizen:
Thanks for merging! 😄 Have a great rest of your day.

Issue #1279, opened 2026-07-29
[1.20.0] `useForm` drops `targetValue` from `lastResult` when `submission.intent` is not a Conform intent


transparent-citizen:
### Describe the bug and the expected behavior

Intents are little confusing to grasp when using React Router with conform, because they refer to one of two things:

1. the intent pattern meant for the server, where you set `name="intent"` and something like `value="post-comment"` or `value="delete-comment"`. On the server, you can later read `request.formData().get("intent")` to figure out which submit button was pressed.
2. Conform's intents, defined via [defineIntent](https://conform.guide/api/react/future/defineIntent). From my understanding, they're supposed to call client-side JavaScript.

In v1.19.4, Conform didn't care when it didn't recognize an intent; it would still apply `lastResult`. With v1.20.0, it drops `targetValue` when the intent is unknown.

### Conform version

v1.20.0

### Steps to Reproduce the Bug or Issue

Run this test:

```tsx
import { report, useForm } from "@conform-to/react/future";
import type { SubmissionResult } from "@conform-to/react/future";
import { render, screen } from "@testing-library/react";
import { createMemoryRouter, Form, RouterProvider } from "react-router";
import { describe, expect, test } from "vitest";
import { z } from "zod";

const testSchema = z.object({
	intent: z.literal("test").default("test"),
	name: z.string(),
});

function renderWithRouter(ui: React.ReactElement) {
	const router = createMemoryRouter([{ path: "/", element: ui }]);
	return render(<RouterProvider router={router} />);
}

function ReproForm({ lastResult }: { lastResult?: SubmissionResult }) {
	const { form, fields } = useForm(testSchema, { lastResult });
	return (
		<Form method="post" id={form.id}>
			<output data-testid="field">
				{String(fields.name.defaultValue ?? "NONE")}
			</output>
		</Form>
	);
}

describe("conform v1.20 unknown intent regression", () => {
	test("unknown intent strips targetValue — passes on 1.19, fails on 1.20+", () => {
		const submission = {
			intent: "test",
			payload: { name: "Alice", intent: "test" },
			fields: ["name", "intent"],
		};

		renderWithRouter(
			<ReproForm
				lastResult={report(submission, { value: { name: "Alice", intent: "test" } })}
			/>,
		);

		expect(screen.getByTestId("field").textContent).toBe("Alice");
	});
});
```

transparent-citizen:
Sorry for the noise, this is slop. `targetValue` has only been available since 1.20.0.

[...]

PR #1280, opened 2026-07-31
feat(conform-react): allow submit and form intents to share a field

edmundhung:
## Summary

- serialize form intents using `type(...)`, including empty parentheses for zero-argument intents
- resolve any value that is not successfully handled as a form intent to a submit intent carrying the original value
- preserve server results for manual submit intents and document the progressive-enhancement pattern
- update intent resolution, SSR initialization, and type-level regression coverage

Related to #1279.

## Verification

- 109 conform-react node tests passed
- TypeScript build and type checks passed
- Oxlint and formatting checks passed

<details>

- Serializes form intents with callable syntax, including zero-argument forms such as `reset()`.
- Treats unrecognized bare values as submit intents with the original payload.
- Keeps malformed or unresolved form intents invalid.
- Supports optional submit payloads and preserves server results during initialization.
- Updates intent parsing, types, documentation, and regression tests.

</details>

### DIFF
--- a/packages/conform-react/future/intent.ts
+++ b/packages/conform-react/future/intent.ts
@@ -69,12 +69,6 @@ export function mergeIntentHandlers<
 
 const undefinedArg = '$$__undefined__$$';
 
-function deserializeIntentArgs(value: string): unknown[] {
-	return JSON.parse(`[${value}]`, (_, value) =>
-		value === undefinedArg ? undefined : value,
-	);
-}
-
 /**
  * Serializes a transport intent to string format.
  */
@@ -86,7 +80,7 @@ export function serializeIntent(intent: TransportIntent): string {
 	}
 
 	if (args.length === 0) {
-		return intent.type;
+		return `${intent.type}()`;
 	}
 
 	const serializedArgs = JSON.stringify(args, function (this, _, value) {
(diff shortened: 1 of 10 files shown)


edmundhung:
@transparent-citizen I think this should avoid the issue you see in #1279. The idea is to treat unrecognized intent as a submit intent instead. That way you could even share the same intent name between your submit intents and form intents.

transparent-citizen:
Thanks for working on this! 🥳 

The differences between form intents and submit intents are raising thoughts about their relation long-term. In a way, they feel like two sides of the same coin. For instance, consider a list of items where users can move an item up:

```tsx
<ul>
    <li>
      <Form {...form.props} method="post">
        <span>Brakes</span>
        <input type="hidden" name="itemId" value="123"/>
        <button type="submit" name="intent" value="move-up">Move up</button>
      </Form>
    </li>
    <li>
      <Form {...form.props} method="post">
        <span>Tubes</span>
        <input type="hidden" name="itemId" value="456"/>
        <button type="submit" name="intent" value="move-up">Move up</button>
      </Form>
    </li>
</ul>
```

Submit intents are guaranteed to be failsafes, since they don't require JS on the client, so forms with these are robust enough to work before hydration and even without script privileges. However, since they cause a submission, moving an item up will add network latency and degrade user experience. As a consequence, users might have to wait a second between clicking "move-up", waiting for the subsequent network roundtrip, and the items eventually rendering in their new order.

Such cases would greatly benefit from a form intent of the same name that Conform may trigger when script execution is available, and whose handler manipulates the item array in client memory. Cutting out the network gives instant feedback.

The best experience would arguably be to gracefully degrade form intents `X()` to submit intents `X`, so that `X()` can run when JS is available, but fall back to `X` when it's not. That relation begs the question if there should be a syntax distinguishing them, or if identically named form/submit intents are useful enough to chain implicitly.

Obviously there are cases where the server doesn't have the capabilities to mirror a form intent (e.g. clipboard interactions), so we can't guarantee that both the server and the client have an intent handler.

I'm curious what your vision with the intent system is!

Regarding form intent syntax, `validate("title")` is not ideal, since there's no way to do static analysis on such strings. People and LLMs will both mess these up. Is the intent name not enough as a discriminator? There is opportunity for type-safety here, since they're all registered through `defineIntent`.

packages/conform-react/future/intent.ts
```diff
@@ -147,32 +142,29 @@ export function parseIntent<
...
+		const handler = options.handlers[transportIntent.type];
 
-	if (!handler) {
-		return undefined;
+		if (handler) {
+			try {
+				return {
+					type: transportIntent.type,
+					payload: handler.parse(...transportIntent.args),
+				} as any;
```
transparent-citizen: Can this type assertion be avoided?

edmundhung: Hmmm, this is a difficult one. 😅 Not sure how to help TypeScript infer this, especially to associate the payload back to the type.

transparent-citizen: I noticed that the type assertion was there even before this PR, so let's not worry about it for now!

edmundhung:
> Submit intents are guaranteed to be failsafes, since they don't require JS on the client, so forms with these are robust enough to work before hydration and even without script privileges. However, since they cause a submission, moving an item up will add network latency and degrade user experience. As a consequence, users might have to wait a second between clicking "move-up", waiting for the subsequent network roundtrip, and the items eventually rendering in their new order.

That's sort of the mental model right now. Conform resolve the intent in the client and decide what to do, and it now basically treat unknown intent as submit intent so it would let the submission reach the server if the form value is valid.
 
> Regarding form intent syntax, `validate("title")` is not ideal, since there's no way to do static analysis on such strings. People and LLMs will both mess these up. Is the intent name not enough as a discriminator? There is opportunity for type-safety here, since they're all registered through `defineIntent`.

It's more of an implementation detail. I don't mean to ask developer to write this themselves. The syntax design is mostly to help debugging as it visualize the intent payload. If you do want to serialize the intent manually for progressive enhancement, my current thought would be an API like this:

```tsx
function Example() {
  const { form, intents } = useForm(...);

  return (
    <form {...form.props}>
       {/* ... */}
       <button
         name={defaultIntentName}
         value={intents.insert.serialize({ name: "todos" })}
       >
         Insert item
       </button>
    </form>
  );
}
```

You can customize the intent name with `configureForms({ intentName: "intent" })` as well.

transparent-citizen:
Understood. Also, LGTM :)

PR #1281, opened 2026-07-31
fix(conform-react): preserve string literal types in DefaultValue

transparent-citizen:
This adds a case to `DefaultValue` so that string literal types are preserved instead of being widened to `string`, making fields like submit intents type-safe.

```ts
const schema = z.object({
	status: z.literal('active'),
	role: z.enum(['admin', 'user']),
	name: z.string(),
});

const form = useForm({
	schema,
	defaultValue: {
		// ❌ before: string | null | undefined
		// ✅  after: 'active' | null | undefined
		status: 'active',

		// ❌ before: string | null | undefined
		// ✅  after: 'admin' | 'user' | null | undefined
		role: 'admin',

		// ✅ string | null | undefined (unchanged)
		name: 'John',
	},
});
```

The PR doesn't touch v1 in anticipation of its soon removal. If this is too soon, please don't hesitate to let me know, and I will retrofit it.

<details>

`DefaultValue` now preserves string literal types in `conform-react/future`, including optional, nullable, template-literal, and nested values. Other types retain existing behavior. Type tests cover these cases and reject incompatible literals. Examples cast URL-derived defaults to schema-specific types. A patch changeset was added.

</details>

### DIFF
--- a/.changeset/olive-trees-narrow.md
+++ b/.changeset/olive-trees-narrow.md
@@ -1,0 +1,5 @@
+---
+'@conform-to/react': patch
+---
+
+Fix `DefaultValue` to preserve string literal types so `z.literal(...)` schemas infer their literal value instead of widening to `string`.
--- a/.changeset/olive-trees-narrow.md
+++ b/.changeset/olive-trees-narrow.md
@@ -1,5 +1,5 @@
 ---
-'@conform-to/react': patch
+'@conform-to/react': minor
 ---
 
 Fix `DefaultValue` to preserve string literal types so `z.literal(...)` schemas infer their literal value instead of widening to `string`.
--- a/examples/headless-ui/src/App.tsx
+++ b/examples/headless-ui/src/App.tsx
@@ -33,6 +33,8 @@ const schema = coerceFormValue(
 	}),
 );
 
+type FormShape = z.input<typeof schema>;
+
 const options = [
 	{ label: 'Durward Reynolds', value: '1' },
 	{ label: 'Kenton Towne', value: '2' },
@@ -56,7 +58,8 @@ export default function App() {
 			color: searchParams.get('color'),
 			project: searchParams.get('project'),
 			notes: searchParams.get('notes'),
-			priority: searchParams.get('priority') ?? 'normal',
+			priority: (searchParams.get('priority') ??
+				'normal') as FormShape['priority'],
 			notifications: searchParams.get('notifications'),
 		},
 		onSubmit(event, { formData, value }) {
--- a/examples/headless-ui/src/App.tsx
+++ b/examples/headless-ui/src/App.tsx
@@ -1,3 +1,4 @@
+import { parseWithZod } from '@conform-to/zod/v4';
 import { coerceFormValue } from '@conform-to/zod/v4/future';
 import {
 	Description,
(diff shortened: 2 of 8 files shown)


transparent-citizen:
The examples are making use of unvalidated search params in `defaultValue`. Since unvalidated search params can contain any `string` and not just allowed literals, tests were failing.

In the real world, we would typically validate all search params before use to prevent unsound application state.

In the examples, I think that a type assertion is good enough to get the point across. By doing that, we're making the promise to TypeScript that these search params will have the correct type, even though TypeScript can show that there are ways for the type to be wider at runtime.

transparent-citizen:
Coderabbit got mad at me for proposing type assertions 😅 So I went and validated all search params.

This should make the examples a little more streamlined, getting rid of the lengthy `defaultValue` objects that read from unvalidated data.

We now have type-safe default values in the examples, so working with string literals is a breeze. This also prevents invalid application state in the examples which was previously possible by constructing URLs that don't validate.

The only downside is that this changes the param semantics in the examples from "use all, might lead to invalid application state" to "use valid only, otherwise start with empty form". I'm not a zod expert, but maybe `.partial` could be useful to improve this in the future.

transparent-citizen:
The changeset was updated from `patch` to `minor` to reflect that users working with loosely-typed data will need to assert types or validate their data, similar to how the PR addresses the examples.

edmundhung:
> In the real world, we would typically validate all search params before use to prevent unsound application state.

I am not sure if that's something Conform can guarantee in the form levels. The default value could also be inferred from `lastResult` where the `submission.payload` could be whatever that being sent to the server. It's also up to the user to adjust the input value from the DOM. I don't see a good way to make sure defaultValue is valid all the time.

In fact, I have been meaning to make `defaultValue` accepts a raw `URLSearchParams` to simplify the example you touched here. Any concerns with #1284?

transparent-citizen:
Hm, I see. The widening seems to have come up as a pain point in v1 before (https://github.com/edmundhung/conform/pull/541), but I understand that `defaultValue` can't make such narrow type guarantees. This could be improved if Conform could validate the schemas internally, which would couple it to every supported validator. 😅

#1284 allows simplifying the examples similarly, without overpromising type guarantees that it can't keep. But it also allows for unsound application state, since malformed URLs can transfer arbitrary data into the application. This is not a regression, it was possible before. Prior validation outside of `useForm`, like this PR does, can prevent that, so in real-world applications, that's safer than passing `URLSearchParams` to `useForm`.

I think #1284 is good to merge nonetheless.

It feels like we're missing a validator-specific layer on top of `useForm` (say `useZodForm`) that helps make stronger guarantees.

transparent-citizen:
Closing. This needs a validation layer somewhere.

edmundhung:
@transparent-citizen I think the narrowing you did is still useful. We can't narrow the output type (e.g. `field.email.defaultValue`) but restricting the `defaultValue` option can still be useful. The type could also be used to teach `useControl()` to only accept the expected enum with `control.change()`.

transparent-citizen:
@edmundhung Just making sure I follow, is the suggestion to give the `defaultValue` option its own narrowed type, as opposed to changing `DefaultValue`?

edmundhung:
@transparent-citizen Changing the `DefaultValue` type as you did is fine. It wouldn't change how we infer the metadata type (e.g. `fields.something.defaultValue` would remain be `string`). 

The case in #541 were an issue only because it impact how we infer `fields.something.value`.

PR #1288, opened 2026-08-03
feat(conform-react): preserve string literal types in DefaultValue

transparent-citizen:
This was previously proposed via #1281, which I had closed prematurely. GitHub doesn't seem to let me reopen it.

Since @edmundhung has kindly migrated the examples to use `URLSearchParams` in the meantime, they shouldn't clash with the narrower `DefaultValue` type like they did in the prior PR.

Description copied over:

This adds a case to `DefaultValue` so that string literal types are preserved instead of being widened to `string`, making fields like submit intents type-safe.

```ts
const schema = z.object({
	status: z.literal('active'),
	role: z.enum(['admin', 'user']),
	name: z.string(),
});

const form = useForm({
	schema,
	defaultValue: {
		// ❌ before: string | null | undefined
		// ✅  after: 'active' | null | undefined
		status: 'active',

		// ❌ before: string | null | undefined
		// ✅  after: 'admin' | 'user' | null | undefined
		role: 'admin',

		// ✅ string | null | undefined (unchanged)
		name: 'John',
	},
});
```

The PR doesn't touch v1 in anticipation of its soon removal. If this is too soon, please don't hesitate to let me know, and I will retrofit it.

<details>

- Preserves string literal and template literal types in `DefaultValue<Shape>`.
- Adds type-level coverage for unions, nullable and optional fields, object properties, and array items.
- Rejects invalid literal defaults at compile time.

</details>

### DIFF
--- a/.changeset/olive-trees-narrow.md
+++ b/.changeset/olive-trees-narrow.md
@@ -1,0 +1,5 @@
+---
+'@conform-to/react': minor
+---
+
+Fix `DefaultValue` to preserve string literal types so `z.literal(...)` schemas infer their literal value instead of widening to `string`.
--- a/packages/conform-react/future/types.ts
+++ b/packages/conform-react/future/types.ts
@@ -195,7 +195,9 @@ export type DefaultValue<Shape> =
 			? Array<DefaultValue<Item>> | null | undefined
 			: Shape extends File | File[]
 				? null | undefined // We don't support setting default value for file inputs yet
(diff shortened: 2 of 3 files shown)


edmundhung (approved):

edmundhung (commented):
Thank you! 😄

transparent-citizen:
My pleasure! Thanks for the quick merge! 🙇🏻

